PR #4, opened 2018-04-12
Put all methods on UserDefaults and remove Defaults

LarsJK:
I'ts not clear to me why you need the Defaults struct?

What do you think about removing it and keeping all methods on UserDefaults instead?

This breaks the API (but could be done without breaking it if preferred..):
- Defaults.Keys is renamed to DefaultsKeys
- Defaults.Key is renamed to DefaultsKey
- Defaults.OptionalKey is renamed to DefaultsOptionalKey
- The global shortcut `defaults` is renamed to `Defaults`.

### DIFF
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -1,27 +1,30 @@
 // MIT License © Sindre Sorhus
 import Foundation
 
-public class DefaultsKeys {
-	fileprivate init() {}
-}
+public extension UserDefaults {
+	public class Keys {
+		fileprivate init() {}
+	}
 
-public final class DefaultsKey<T: Codable>: DefaultsKeys {
-	fileprivate let name: String
-	fileprivate let defaultValue: T
+	public final class Key<T: Codable>: Keys {
+		fileprivate let name: String
+		fileprivate let defaultValue: T
 
-	public init(_ key: String, default defaultValue: T) {
-		self.name = key
-		self.defaultValue = defaultValue
+		public init(_ key: String, default defaultValue: T) {
+			self.name = key
+			self.defaultValue = defaultValue
+		}
+	}
+
+	public final class OptionalKey<T: Codable>: Keys {
+		fileprivate let name: String
+
+		public init(_ key: String) {
+			self.name = key
+		}
 	}
 }
 
-public final class DefaultsOptionalKey<T: Codable>: DefaultsKeys {
-	fileprivate let name: String
-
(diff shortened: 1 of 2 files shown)


LarsJK:
I agree that that Keys types should be nested. They could be nested onto UserDefaults in an extension (didn't actually know I could do that..)

I see your point about not being tied to UserDefaults.
The autocomplete spamming could be solved by adding a protocol.

I dont have a strong opinion about this, but here are the benefits I see:

- I think the Defaults struct would be hard to fake when testing.
- It would be easier to adopt because I wouldn't have to change any interfaces, but I see that most of the methods are available on UserDefaults as well, so I guess it wouldn't make a big difference..
- An extension can be used with other extensions or subclasses of UserDefaults.

sindresorhus:
> They could be nested onto UserDefaults in an extension (didn't actually know I could do that..)

I had that too at some point. I've rewritten this module so many times... The reason I moved away from that is that it's weird having to put keys in `UserDefaults.Keys` while you use `Defaults[.foo]`.

> I think the Defaults struct would be hard to fake when testing.

Why would it be hard to fake? And why would you need to fake it? It's just UserDefaults underneath.

> It would be easier to adopt because I wouldn't have to change any interfaces, but I see that most of the methods are available on UserDefaults as well, so I guess it wouldn't make a big difference..

I'd rather optimize for a great API than for making easy to move existing code to.

> An extension can be used with other extensions or subclasses of UserDefaults.

An extension can also clash with other extensions. I have never seen anyone subclassing UserDefaults.

LarsJK:
This is what I usually do with UserDefaults:

```swift
func myFunc(defaults: UserDefaults = .standard) -> String? {
	return defaults.string(forKey: "my-key")
}

// In tests
class InMemoryUserDefaults: UserDefaults {
	var defaults: [String: Any] = [:]

	override func string(forKey defaultName: String) -> String? {
		return defaults[defaultName] as? String
	}
}

let fakeDefaults = InMemoryUserDefaults()
fakeDefaults.defaults["my-key"] = "Hello"

myFunc(defaults: fakeDefaults)
```

You could do something similar with Defaults if you add a protocol to it, and use that in your function tho..

I understand your reasoning now, so I will just close this PR..

sindresorhus:
I cherry-picked some of your unrelated changes in https://github.com/sindresorhus/Defaults/commit/3dffe718a2849680124d5488e7b9d3baedd21a89 Thanks! :)

PR #10, opened 2018-10-03
Add support for key observation

sindresorhus:
And some other improvements:
- Shorter syntax for defining keys.
- Support alternative UserDefaults suites.
- Improved docs and test coverage.

### DIFF
--- a/Sources/Defaults.swift
+++ b/Sources/Defaults.swift
@@ -3,48 +3,62 @@ import Foundation
 
 public final class Defaults {
 	public class Keys {
+		public typealias Key = Defaults.Key
+		public typealias OptionalKey = Defaults.OptionalKey
+
 		fileprivate init() {}
 	}
 
 	public final class Key<T: Codable>: Keys {
-		fileprivate let name: String
-		fileprivate let defaultValue: T
+		public let name: String
+		public let defaultValue: T
+		public let suite: UserDefaults
 
-		public init(_ key: String, default defaultValue: T) {
+		public init(_ key: String, default defaultValue: T, suite: UserDefaults = .standard) {
 			self.name = key
 			self.defaultValue = defaultValue
+			self.suite = suite
+
+			super.init()
+
+			// Sets the default value in the actual user defaults so it can be used in other contexts, like binding.
+			suite[self] = defaultValue
 		}
 	}
 
 	public final class OptionalKey<T: Codable>: Keys {
-		fileprivate let name: String
+		public let name: String
+		public let suite: UserDefaults
 
-		public init(_ key: String) {
+		public init(_ key: String, suite: UserDefaults = .standard) {
 			self.name = key
+			self.suite = suite
 		}
 	}
 
+	fileprivate init() {}
+
 	public subscript<T: Codable>(key: Defaults.Key<T>) -> T {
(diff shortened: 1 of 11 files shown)


sindresorhus:
@LarsJK @LinusU If you have time, I would appreciate a review. You can look at the readme changes for what's being added.

readme.md
```diff
@@ -98,76 +98,152 @@ enum DurationKeys: String, Codable {
...
 
 Clear the user defaults.
 
+#### `Defaults#observe`
+
+```swift
+observe<T: Codable>(
+	_ key: Defaults.Key<T>,
+	options: NSKeyValueObservingOptions = [.initial, .old, .new],
+	handler: @escaping (OptionalKeyChange<T>) -> Void
```
LarsJK: Should be KeyChange not OptionalKeyChange

readme.md
```diff
@@ -98,76 +98,152 @@ enum DurationKeys: String, Codable {
...
 
-#### Defaults.Keys
+#### `Defaults.Keys`
 
 Type: `class`
 
 Stores the keys.
 
-#### Defaults.Key
+#### `Defaults.Key` *(alias `Defaults.Keys.OptionalKey`)*
```
LarsJK: should be alias `Defaults.Keys.Key` not OptionalKey

readme.md
```diff
@@ -98,76 +98,152 @@ enum DurationKeys: String, Codable {
...
 
 Create a key with a default value.
 
-#### Defaults.OptionalKey
+The default value is written to the actual `UserDefaults` and can be used elsewhere. For example, with Interface Builder binding.
+
+#### `Defaults.OptionalKey` *(alias `Defaults.Keys.OptionalKey`)*
+
+```swift
+Defaults.OptionalKey<T>(_ key: String, default: T)
```
LarsJK: Has no default value and is missing the suite parameter

Sources/Defaults.swift
```diff
@@ -115,7 +129,7 @@ extension UserDefaults {
 		}
 	}
 
-	private func isNativelySupportedType<T>(_ type: T.Type) -> Bool {
+	fileprivate static func isNativelySupportedType<T>(_ type: T.Type) -> Bool {
```
LarsJK: I think I would just leave it as a private method but it doesn't really make a difference..

sindresorhus: Good catch. In an earlier iteration, I used it in the Observation file, and I forgot to change it back to private.

readme.md
sindresorhus: I should not copy-paste stuff when I'm sleepy 😆

sindresorhus:
I have absolutely no idea why the observer closure is being called twice on Travis, but not locally:

```
DefaultsTests.swift:123: error: -[Defaults_macOS_Tests.DefaultsTests testObserveKey] : failed: caught "NSInternalInconsistencyException", "API violation - multiple calls made to -[XCTestExpectation fulfill] for Observation closure being called."
```

Any ideas?

sindresorhus:
Alright. I managed to fix the tests on Travis in the end.

sindresorhus:
Thanks for reviewing ❤️

PR #12, opened 2018-11-01
Update readme.md w.r.t. SwiftyUserDefaults

vanshg:
According to https://github.com/radex/SwiftyUserDefaults#codable, Codable support exists in SwiftyUserDefaults since v4.


sindresorhus:
Codable support is only in 4.0.0-alpha.1, so not really usable for anyone. I'll update when it's in a stable release.

vanshg:
Gotcha

PR #13, opened 2019-03-24
Add `DefaultsObsevation.tieToLifetime(of:)`

ThatsJustCheesy:
Strongly associates the observation with the passed-in object. It then lives for at least as long as the object, and automatically invalidates either a) once no more references to it exist, or b) when it receives a notification and realizes that the object to which it was tied has died.

### DIFF
--- a/Sources/Defaults/Observation.swift
+++ b/Sources/Defaults/Observation.swift
@@ -25,21 +25,6 @@ public protocol DefaultsObservation: AnyObject {
 	func removeLifetimeTie()
 }
 
-private struct DefaultsObservationAssociationKeys {
-	static let lifetimeAssociation = AssociatedObject<LifetimeAssociation>()
-}
-
-extension DefaultsObservation {
-	public func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
-		DefaultsObservationAssociationKeys.lifetimeAssociation[self] = associate(self, with: weaklyHeldObject)
-		return self
-	}
-	
-	public func removeLifetimeTie() {
-		DefaultsObservationAssociationKeys.lifetimeAssociation[self] = nil
-	}
-}
-
 extension Defaults {
 	private static func deserialize<T: Decodable>(_ value: Any?, to type: T.Type) -> T? {
 		guard
@@ -130,7 +115,20 @@ extension Defaults {
 		public func invalidate() {
 			object?.removeObserver(self, forKeyPath: key, context: nil)
 			object = nil
-			DefaultsObservationAssociationKeys.lifetimeAssociation[self] = nil
+			lifetimeAssociation?.cancel()
+		}
+
+		private var lifetimeAssociation: LifetimeAssociation? = nil
+
+		func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
+			lifetimeAssociation = LifetimeAssociation(of: self, with: weaklyHeldObject, deinitHandler: { [weak self] in
+				self?.invalidate()
+			})
+			return self
+		}
+
+		func removeLifetimeTie() {
+			lifetimeAssociation?.cancel()
 		}
 
 		// swiftlint:disable:next block_based_kvo
@@ -149,13 +147,6 @@ extension Defaults {
 				let change = change
 			else {
 				return
-			}
-
-			if let lifetimeAssociation = DefaultsObservationAssociationKeys.lifetimeAssociation[self] {
-				guard lifetimeAssociation.isValid else {
-					invalidate()
-					return
-				}
 			}
 
 			callback(BaseChange(change: change))
--- a/Sources/Defaults/Observation.swift
+++ b/Sources/Defaults/Observation.swift
@@ -4,8 +4,22 @@ import Foundation
 public protocol DefaultsObservation: AnyObject {
 	func invalidate()
 
+	/**
+	Keep this observation alive for as long as, and no longer than, another
+	object exists.
+
+	```
+	defaults.observe(.xyz) { [unowned self] change in
+		self.xyz = change.newValue
+	}.tieToLifetime(of: self)
+	```
+	*/
 	@discardableResult
 	func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self
+	/**
+	Break the lifetime tie created by `tieToLifetime(of:)`, if one exists.
+	- Postcondition: As if `tieToLifetime(of:)` was never called on self.
+	*/
 	func removeLifetimeTie()
 }
 
--- a/Sources/Defaults/Observation.swift
+++ b/Sources/Defaults/Observation.swift
@@ -1,11 +1,11 @@
 import Foundation
 
 /// TODO: Nest this inside `Defaults` if Swift ever supported nested protocols.
-public protocol DefaultsObservation: NSObjectProtocol {
+public protocol DefaultsObservation: AnyObject {
 	func invalidate()
 
 	@discardableResult
-	func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> DefaultsObservation
+	func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self
 	func removeLifetimeTie()
 }
 
@@ -109,7 +109,7 @@ extension Defaults {
 		private var lifetimeTie: LifetimeTie?
 		private static var lifetimeTieAssociationKey = AssociatedObject<UserDefaultsKeyObservation>()
 
-		public func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> DefaultsObservation {
+		public func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
 			UserDefaultsKeyObservation.lifetimeTieAssociationKey[weaklyHeldObject] = self
 			lifetimeTie = LifetimeTie(object: weaklyHeldObject)
 			return self
--- a/Sources/Defaults/Observation.swift
+++ b/Sources/Defaults/Observation.swift
@@ -9,6 +9,21 @@ public protocol DefaultsObservation: AnyObject {
 	func removeLifetimeTie()
 }
 
+private struct DefaultsObservationAssociationKeys {
+	static let lifetimeAssociation = AssociatedObject<LifetimeAssociation>()
+}
+
+extension DefaultsObservation {
+	public func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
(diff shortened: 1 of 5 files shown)


ThatsJustCheesy:
Think I'll rename it to `tieToLifetime(of:)`.

I also considered `tieTo` → `tyingTo`, but that's typically for methods that are absolutely nonmutating; this mutates its receiver and returns it as a convenience.

Sources/Observation.swift
```diff
@@ -1,8 +1,12 @@
 import Foundation
 
 /// TODO: Nest this inside `Defaults` if Swift ever supported nested protocols.
-public protocol DefaultsObservation {
+public protocol DefaultsObservation: NSObjectProtocol {
```
sindresorhus:
```suggestion
public protocol DefaultsObservation: AnyObject {
```

Sources/Observation.swift
```diff
@@ -95,6 +99,30 @@ extension Defaults {
...
+
+		private var lifetimeTie: LifetimeTie?
+		private static var lifetimeTieAssociationKey = AssociatedObject<UserDefaultsKeyObservation>()
+
+		public func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> DefaultsObservation {
+			UserDefaultsKeyObservation.lifetimeTieAssociationKey[weaklyHeldObject] = self
+			lifetimeTie = LifetimeTie(object: weaklyHeldObject)
+			return self
+		}
+
```
sindresorhus: Any way we could make the lifetime tying logic reusable? So we could use it for other things too, not just in this module.


Something like this:

```swift
private let bindLifetimeAssociatedObjectKey = ObjectAssociation<[AnyObject]>()

func bindLifetime(of object: AnyObject, to target: AnyObject) {
	var retainedObjects = bindLifetimeAssociatedObjectKey[target] ?? []
	retainedObjects.append(object)
	bindLifetimeAssociatedObjectKey[target] = retainedObjects
}
```

But `target` should be weakly held and maybe we should return an untie function from it.

And any thoughts of the use of `bind` vs `tie`?

We also need a way for it to cleanup, for example, in this case, we need to call `.invalidate()`. Maybe a `deinit` closure argument?

Sources/Observation.swift
```diff
@@ -1,8 +1,12 @@
 import Foundation
 
 /// TODO: Nest this inside `Defaults` if Swift ever supported nested protocols.
-public protocol DefaultsObservation {
+public protocol DefaultsObservation: NSObjectProtocol {
 	func invalidate()
+
+	@discardableResult
+	func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> DefaultsObservation
```
sindresorhus:
```suggestion
	func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> Self
```

sindresorhus: This method could use a doc comment.

Sources/Observation.swift
```diff
@@ -1,8 +1,12 @@
...
+	func removeLifetimeTie()
```
sindresorhus: This method could use a doc comment.

Sources/Observation.swift
```diff
@@ -95,6 +99,30 @@ extension Defaults {
...
+		}
+
+		private struct LifetimeTie {
+			weak var object: AnyObject?
+		}
+
+		private var lifetimeTie: LifetimeTie?
+		private static var lifetimeTieAssociationKey = AssociatedObject<UserDefaultsKeyObservation>()
+
+		public func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> DefaultsObservation {
```
sindresorhus: What do you think of implementation these two methods as default implementations of the protocol instead?

Sources/Observation.swift
```diff
@@ -95,6 +99,30 @@ extension Defaults {
...
+			UserDefaultsKeyObservation.lifetimeTieAssociationKey[weaklyHeldObject] = self
+			lifetimeTie = LifetimeTie(object: weaklyHeldObject)
+			return self
+		}
+
+		public func removeLifetimeTie() {
```
sindresorhus: This also needs a test.

Sources/Observation.swift
sindresorhus:
```suggestion
		public func tieToLifetimeOf(_ weaklyHeldObject: AnyObject) -> Self {
```

Sources/Observation.swift
ThatsJustCheesy: I tried but couldn't get ARC to reliably kill the object during the test's lifetime. I tried an autoreleasepool block, waiting, etc. but the test kept working locally and failing in the Travis build. I can keep trying though.

ThatsJustCheesy: Never mind, I was overthinking it. :D

Just wait a while and test a weak reference for `nil`.

ThatsJustCheesy:
I've generalized the association logic. This also makes for a nicer internal interface, IMO.

Meanwhile, the logic for invalidation when the owner disappears is left up to the specific user class. This way, the infrastructure can be used to simply extend an object's lifetime, rather than bind it to die when its owner dies. Also, there's no good way to proactively determine when a weak reference goes nil. So instead, the association object exposes `var isValid: Bool` that, if desired, a user can check for falsehood and then invalidate itself as appropriate.

sindresorhus:
> Also, there's no good way to proactively determine when a weak reference goes nil.

https://stackoverflow.com/questions/28670796/can-i-hook-when-a-weakly-referenced-object-of-arbitrary-type-is-freed?noredirect=1&lq=1

Sources/util.swift
```diff
@@ -27,3 +27,40 @@ final class AssociatedObject<T: Any> {
 		}
 	}
 }
+
+class LifetimeAssociation {
```
sindresorhus: Could use a doc comment.

Sources/util.swift
```diff
@@ -27,3 +27,40 @@ final class AssociatedObject<T: Any> {
...
+		associatedObjects.remove(at: objectIndex)
+		LifetimeAssociationKeys.associatedObjects[owner] = associatedObjects
+	}
+}
+
+private struct LifetimeAssociationKeys {
+	static let associatedObjects = AssociatedObject<[AnyObject]>()
+}
+
+func associate(_ object: AnyObject, with owner: AnyObject) -> LifetimeAssociation {
```
sindresorhus: Could use a doc comment with an example.

Sources/Observation.swift
```diff
@@ -9,6 +9,21 @@ public protocol DefaultsObservation: AnyObject {
 	func removeLifetimeTie()
 }
 
+private struct DefaultsObservationAssociationKeys {
+	static let lifetimeAssociation = AssociatedObject<LifetimeAssociation>()
+}
```
sindresorhus: I was hoping abstracting the lifetime handling to `LifetimeAssociation` would let us not have to use associated object here.

Sources/util.swift
```diff
@@ -27,3 +27,40 @@ final class AssociatedObject<T: Any> {
 		}
 	}
 }
+
+class LifetimeAssociation {
+	private(set) weak var object: AnyObject?
+	private(set) weak var owner: AnyObject?
```
sindresorhus: Why are these `private(set)` and not `private`? I don't see them being read anywhere outside the class.

Sources/util.swift
```diff
@@ -27,3 +27,40 @@ final class AssociatedObject<T: Any> {
...
+		else {
+			return
+		}
+		associatedObjects.remove(at: objectIndex)
+		LifetimeAssociationKeys.associatedObjects[owner] = associatedObjects
+	}
+}
+
+private struct LifetimeAssociationKeys {
+	static let associatedObjects = AssociatedObject<[AnyObject]>()
```
sindresorhus: This could just be in `LifetimeAssociation` instead.

Sources/util.swift
sindresorhus:
```suggestion
final class LifetimeAssociation {
```

Sources/util.swift
sindresorhus: Is the naming `associate` clear enough that it associates the lifetime? Maybe `associateLifetime(of:with:)`?

sindresorhus:
```suggestion
@discardableResult
func associate(_ object: AnyObject, with owner: AnyObject) -> LifetimeAssociation {
```

for cases where you don't need to do anything with the association?

sindresorhus:
@ThatsJustCheesy Ping :)

sindresorhus:
@ThatsJustCheesy Bump :)

ThatsJustCheesy:
I've refined the design further. We've now got a pretty reusable LifetimeAssociation class, and a couple of methods on DefaultsObservation/UserDefaultsKeyObservation that use it.

Apologies for leaving this for quite some time; as I'm sure you know, this sort of existential memory management logic can be hard to reason about :D

Sources/Defaults/Observation.swift
```diff
@@ -95,6 +115,20 @@ extension Defaults {
 		public func invalidate() {
 			object?.removeObserver(self, forKeyPath: key, context: nil)
 			object = nil
+			lifetimeAssociation?.cancel()
+		}
+
+		private var lifetimeAssociation: LifetimeAssociation? = nil
+
+		func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
```
sindresorhus:
```suggestion
		public func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
```

Sources/Defaults/Observation.swift
```diff
@@ -95,6 +115,20 @@ extension Defaults {
...
+			lifetimeAssociation = LifetimeAssociation(of: self, with: weaklyHeldObject, deinitHandler: { [weak self] in
+				self?.invalidate()
+			})
+			return self
+		}
+
+		func removeLifetimeTie() {
```
sindresorhus:
```suggestion
		public func removeLifetimeTie() {
```

sindresorhus:
I'm very happy about how this turned out in the end. Really nice work. 👌

sindresorhus:
Can you add a small example to the readme and mention the two methods in the API section? (doesn't have to be comprehensive, just so users know it exists

ThatsJustCheesy:
Should be good to go.

sindresorhus:
Yay! Thanks for persevering through this. It's a really useful addition. 🙌

PR #21, opened 2019-08-08
New feature to clear only specific keys

Maschina:
This pull requests adds the feature to reset only single keys back to their default value. For optional keys, the default value will be `nil`.

### DIFF
--- a/Configs/Defaults.plist
+++ b/Configs/Defaults.plist
@@ -13,7 +13,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.0.2</string>
+	<string>$(MARKETING_VERSION)</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>NSHumanReadableCopyright</key>
--- a/Configs/Defaults.plist
+++ b/Configs/Defaults.plist
@@ -13,7 +13,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>$(MARKETING_VERSION)</string>
+	<string>2.0.2</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>NSHumanReadableCopyright</key>
--- a/Defaults.podspec
+++ b/Defaults.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
 	s.name = 'Defaults'
-	s.version = '2.0.2'
+	s.version = '2.1'
 	s.summary = 'Swifty and modern UserDefaults'
 	s.license = 'MIT'
 	s.homepage = 'https://github.com/sindresorhus/Defaults'
--- a/Defaults.podspec
+++ b/Defaults.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
 	s.name = 'Defaults'
-	s.version = '2.1'
+	s.version = '2.0.2'
 	s.summary = 'Swifty and modern UserDefaults'
 	s.license = 'MIT'
 	s.homepage = 'https://github.com/sindresorhus/Defaults'
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -65,6 +65,18 @@ public final class Defaults {
 			suite.removeObject(forKey: key)
 		}
 	}
+	
+	public func clear<T: Codable>(key: Defaults.Key<T>, suite: UserDefaults = .standard) {
+		for key in suite.dictionaryRepresentation().keys.filter({ $0 == key.name }) {
+			suite.removeObject(forKey: key)
+		}
+	}
+	
+	public func clear<T: Codable>(key: Defaults.OptionalKey<T>, suite: UserDefaults = .standard) {
+		for key in suite.dictionaryRepresentation().keys.filter({ $0 == key.name }) {
+			suite.removeObject(forKey: key)
+		}
+	}
 }
 
 // Has to be `defaults` lowercase until Swift supports static subscripts…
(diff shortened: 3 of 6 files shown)


Defaults.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
 	s.name = 'Defaults'
-	s.version = '2.0.2'
+	s.version = '2.1'
```
sindresorhus: Don’t change the version. I will do so when publishing the new version.

Sources/Defaults/Defaults.swift
```diff
@@ -65,6 +65,18 @@ public final class Defaults {
 			suite.removeObject(forKey: key)
 		}
 	}
+	
+	public func clear<T: Codable>(key: Defaults.Key<T>, suite: UserDefaults = .standard) {
+		for key in suite.dictionaryRepresentation().keys.filter({ $0 == key.name }) {
```
sindresorhus: Use the “where” keyword instead of “filter”.

Sources/Defaults/Defaults.swift
```diff
@@ -65,6 +65,18 @@ public final class Defaults {
 			suite.removeObject(forKey: key)
 		}
 	}
+	
+	public func clear<T: Codable>(key: Defaults.Key<T>, suite: UserDefaults = .standard) {
```
sindresorhus: Can you add a documentation comment for this and the other clear method?

Sources/Defaults/Defaults.swift
```diff
@@ -65,6 +65,18 @@ public final class Defaults {
...
+		for key in suite.dictionaryRepresentation().keys.filter({ $0 == key.name }) {
+			suite.removeObject(forKey: key)
+		}
+	}
+	
+	public func clear<T: Codable>(key: Defaults.OptionalKey<T>, suite: UserDefaults = .standard) {
```
sindresorhus:
```suggestion
	public func clear<T: Codable>(_ key: Defaults.OptionalKey<T>, suite: UserDefaults = .standard) {
```

Defaults.podspec
Maschina:
```suggestion
	s.version = '2.0.2'
```

Sources/Defaults/Defaults.swift
Maschina: I have to apologize but I am not 100% sure what you mean. Were you thinking of `first(where: ...`?

Sources/Defaults/Defaults.swift
sindresorhus: I think it would be useful to accept multiple keys.

Sources/Defaults/Defaults.swift
```diff
@@ -65,6 +65,18 @@ public final class Defaults {
 			suite.removeObject(forKey: key)
 		}
 	}
+	
+	public func clear<T: Codable>(key: Defaults.Key<T>, suite: UserDefaults = .standard) {
+		for key in suite.dictionaryRepresentation().keys.filter({ $0 == key.name }) {
+			suite.removeObject(forKey: key)
+		}
```
sindresorhus: Why not just:

```swift
for key in keys {
	self[key] = key.defaultValue
}
```

?

Sources/Defaults/Defaults.swift
Maschina: Sure, documentation is useful here.

Configs/Defaults.plist
```diff
@@ -13,7 +13,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>2.0.2</string>
+	<string>$(MARKETING_VERSION)</string>
```
sindresorhus: Can you remove all the other unrelated changes like this one?

Sources/Defaults/Defaults.swift
Maschina: > I think it would be useful to accept multiple keys.

I agree, but I would rather implement it in an additional method.
Something like this: `public func clear<T: Codable>(keys: [Defaults.Key<T>]`

What's your opinion?

sindresorhus:
This is what I currently use in one of my projects:

```swift
extension Defaults {
	/// Reset `Defaults.Key` items back to their default value.
	func reset<T: Codable>(_ keys: Defaults.Key<T>...) {
		reset(keys)
	}

	/// Reset `Defaults.Key` items back to their default value.
	func reset<T: Codable>(_ keys: [Defaults.Key<T>]) {
		for key in keys {
			self[key] = key.defaultValue
		}
	}

	/// Reset `Defaults.OptionalKey` items back to nil.
	func reset<T: Codable>(_ keys: Defaults.OptionalKey<T>...) {
		reset(keys)
	}

	/// Reset `Defaults.OptionalKey` items back to nil.
	func reset<T: Codable>(_ keys: [Defaults.OptionalKey<T>]) {
		for key in keys {
			self[key] = nil
		}
	}
}
```

Sources/Defaults/Defaults.swift
Maschina: Good idea. Would this also work for `OptionalKey`?

Configs/Defaults.plist
Maschina:
```suggestion
	<string>2.0.2</string>
```

Sources/Defaults/Defaults.swift
sindresorhus: Yes: https://github.com/sindresorhus/Defaults/pull/21#issuecomment-519684656

Maschina:
Nice code. Let me adapt it accordingly.

sindresorhus:
I think I also prefer the `reset` name as it more accurately describe what it does; it resets back to the default value.

Sources/Defaults/Defaults.swift
sindresorhus: Do it like https://github.com/sindresorhus/Defaults/pull/21#issuecomment-519684656

Sources/Defaults/Defaults.swift
sindresorhus:
```swift
for suiteKey in suite.dictionaryRepresentation().keys where suiteKey == key {
```

PR #22, opened 2019-08-08
Add `.reset()` method to reset given keys back to their default value

Maschina:
This pull requests adds the feature to reset only single keys back to their default value. For optional keys, the default value will be nil.

Continued from https://github.com/sindresorhus/Defaults/pull/21.


sindresorhus:
You need to document it in the API section too: https://github.com/sindresorhus/Defaults#api

sindresorhus:
One concern is that it now might be unclear what the difference between the existing `.clear()` method and the `reset()` method is. Maybe we should renamed `.clear()` to `.clearAll()` to make it clear (pun intended)?

sindresorhus:
@Maschina Bump :)

PR #27, opened 2019-10-06
Fix #16 NSSecureCoding support

### LINKED ISSUE #16: NSSecureCoding/NSCoding support
Hi
Defaults is awesome lib to manage UserDefaults in Swifty style. Unfortunately, it does not support NSSecureCoding protocol right now. It is useful to work with some Apple-specific classes which support NSSecureCoding only. For example, take a look at `NSPersistentHistoryToken`.

Ybrin:
This fixes #16 

I didn't want to make any breaking changes so I just added `NSSecureCodingKey` and `NSSecureCodingOptionalKey` which can be used just like `Key` and `OptionalKey`.

Observing them also works just like with the Codables.

I added tests for all of the new methods according to the other tests.

### DIFF
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -4,7 +4,14 @@ import Foundation
 public final class Defaults {
 	public class Keys {
 		public typealias Key = Defaults.Key
+
+		@available(iOSApplicationExtension 11.0, *)
+		public typealias NSSecureCodingKey = Defaults.NSSecureCodingKey
+
 		public typealias OptionalKey = Defaults.OptionalKey
+
+		@available(iOSApplicationExtension 11.0, *)
+		public typealias NSSecureCodingOptionalKey = Defaults.NSSecureCodingOptionalKey
 
 		fileprivate init() {}
 	}
@@ -31,7 +38,42 @@ public final class Defaults {
 		}
 	}
 
+	@available(iOSApplicationExtension 11.0, *)
+	public final class NSSecureCodingKey<T: NSSecureCoding>: Keys {
+		public let name: String
+		public let defaultValue: T
+		public let suite: UserDefaults
+
+		/// Create a defaults key.
+		public init(_ key: String, default defaultValue: T, suite: UserDefaults = .standard) {
+			self.name = key
+			self.defaultValue = defaultValue
+			self.suite = suite
+
+			super.init()
+
+			// Sets the default value in the actual UserDefaults, so it can be used in other contexts, like binding.
+			if UserDefaults.isNativelySupportedType(T.self) {
+				suite.register(defaults: [key: defaultValue])
+			} else if let value = try? NSKeyedArchiver.archivedData(withRootObject: defaultValue, requiringSecureCoding: true) {
+				suite.register(defaults: [key: value])
+			}
+		}
+	}
+
 	public final class OptionalKey<T: Codable>: Keys {
+		public let name: String
+		public let suite: UserDefaults
+
+		/// Create an optional defaults key.
+		public init(_ key: String, suite: UserDefaults = .standard) {
+			self.name = key
+			self.suite = suite
+		}
+	}
+
+	@available(iOSApplicationExtension 11.0, *)
+	public final class NSSecureCodingOptionalKey<T: NSSecureCoding>: Keys {
 		public let name: String
 		public let suite: UserDefaults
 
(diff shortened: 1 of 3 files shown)


Ybrin:
@sindresorhus

sindresorhus:
Awesome 👍

sindresorhus:
You also need to document it in the readme. (highlights section and API docs).

Seems like the tests are failing.

Ybrin:
@sindresorhus 

Tests are passing now. I forgot to support macOS, tvOS and watchOS before.

I added some examples to the readme and extended all API docs.

Do you want me to bump the podspec version too?

sindresorhus:
Is there really no way of just using the existing `Key` and `OptionalKey` types instead of duplicating everything?

Ybrin:
@sindresorhus I don't see any non breaking change I could make to use `Key` and `OptionalKey`.

We somehow have to maintain the generic parameter in the Key types and also make sure they still accept `Codable`. But at the same time they have to accept `NSSecureCoding`. This would only be possible if we made `NSSecureCoding` conform to `Codable`, which is not allowed as both are protocols.    
Theoretically users could always just implement `Codable` into their `NSSecureCoding` classes without us having to explicitly support it. But we are trying to make it a little bit easier here for people who don't need `Codable` support in their types.

The main issue is that `Codable` and `NSSecureCoding` are not related in any way. We can't dynamically convert one to the other during runtime. Because of that even the breaking change would require us to completely rethink the way Defaults is used. And I don't think we should do this as I really like the API as it currently is.    
But I am open to ideas if you have any.

Maybe @sunshinejr

sindresorhus:
// @ThatsJustCheesy @SergeyKuryanov In case you have any ideas.

sunshinejr:
Can you add an additional init to the `Key` and `OptionalKey`?
```swift
init<V: NSSecureCoding>(_ key: String, default defaultValue: T & V, suite: UserDefaults = .standard) {

}
```

Ybrin:
@sunshinejr Is it possible not to use both generic parameters? I think the issue here is that if you pass a `NSSecureCoding` parameter and the initializer only uses `V`, the compiler doesn't know which type to use for `T` and will complain, even if you don't use the `T` parameters anywhere else.

sunshinejr:
@Ybrin ah yeah you're probably right, in that case maybe just make `V: NSSecureCoding, V: Codable`?

Ybrin:
@sunshinejr But where should we use the `Codable` type? If it is really `defaultValue: T & V ` we would require a type which conforms to both, `Codable` and `NSSecureCoding`. A second initializer would always mean we need to get a `Codable` type from the caller without using it (setting to nil afterwards). Also the other way round with the `Codable` initializer.

Here is an example

```

**Comment by Ybrin** [comment]
@sunshinejr But where should we use the `Codable` type? If it is really `defaultValue: T & V` we would require a type which conforms to both, `Codable` and `NSSecureCoding`. A second initializer would always mean we need to get a `Codable` type from the caller without using it (setting to nil afterwards). Also the other way round with the Codable initializer.

Here is an example

```swift
public final class Key<T: Codable, V: NSSecureCoding>: Keys {
	public let name: String
	public let defaultValueCodable: T?
	public let defaultValueCoding: V?
	public let suite: UserDefaults

	/// Create a Codable defaults key.
	public init(_ key: String, default defaultValue: T, suite: UserDefaults = .standard, secureCodingType: V.Type = String.self) {
		self.name = key
		self.defaultValueCodable = defaultValue
		self.defaultValueCoding = nil
		// ... Setup
	}

	/// Create a NSSecureCoding defaults key.
	public init(_ key: String, default defaultValueCoding: V, suite: UserDefaults = .standard, codableType: T.Type = SomeNSSecureCodingClass.self) {
		self.name = key
		self.defaultValueCoding = defaultValue
		self.defaultValueCodable = nil
		// ... Setup
	}
}
```

I guess I could do it that way (haven't tested it yet whether the compiler complains or not) but it seems a little bit hacky...    
I can do it though, if it's okay like that for you...

Ybrin:
tl;dr: I am using a default value for a pointless parameter in both initializers which determines the type of the generic parameter we are not using and then set it to nil. This way the API doesn't break as there is the default value for it and the compiler still has everything it needs.

The API just seems a little bit weird as we would have to ask the users to just ignore the last parameter of this initializers...

Ybrin:
Btw, it would break a little bit for people using explicit types when initializing `Key` or `OptionalKey`. I don't really think anybody does this, but just as a side note...

Ybrin:
The only real issue I see with this approach is that we would have to make the whole API only available after iOS 11.0, macOS 10.13, tvOS 11.0 and watchOS 4.0 because we would use `NSSecureCoding` as a generic parameter in both `Key` and `OptionalKey`, which are used throughout the whole API.

What do you think @sindresorhus ? Should I go for it?

sindresorhus:
> Btw, it would break a little bit for people using explicit types when initializing Key or OptionalKey

That's not something I want to break as it's the recommended way to use `Defaults`.

> tl;dr: I am using a default value for a pointless parameter in both initializers which determines the type of the generic parameter we are not using and then set it to nil.

That is quite ugly indeed...

sindresorhus:
I think duplicating the code (the current state of the PR) is better than what's proposed yet, but I'll leave this open for some more days in case someone has a better idea on how to reduce the duplication.

Ybrin:
@ThatsJustCheesy @SergeyKuryanov Do you have any ideas? I thought about it again. And because NSSecureCoding and Codable are not related in any way, duplicating everything is the right option, at least from a clean design standpoint. I am still open to ideas though...

Issue #53, opened 2020-06-10
Support `[String: String]` as a type


sindresorhus:
Any string-keyed dictionary with the value type of any of the supported types.

Will either require https://github.com/sindresorhus/Defaults/issues/52 or some [built-in custom handling](https://www.vadimbulavin.com/advanced-guide-to-userdefaults-in-swift/).

sindresorhus:
Fixed by #57

Issue #51, opened 2020-06-10
Improve check for natively supported values


sindresorhus:
https://github.com/apple/swift-corelibs-foundation/blob/ef6f96ee82ea0f54252071c0ecadf5f01be9aecc/Foundation/UserDefaults.swift#L26-L64

Seems we could also support `NSNumber` and various `Int*` types.

Even arrays of values (https://github.com/sindresorhus/Defaults/issues/47).

We could also adopt the approach of using a protocol to define the natively supported types: https://www.vadimbulavin.com/advanced-guide-to-userdefaults-in-swift/

sindresorhus:
Fixed by #57

PR #57, opened 2020-12-21
Add a `Defaults.Serializable` protocol

### LINKED ISSUE #52: Add a `DefaultsSerializable` protocol to make it possible to do custom serializations/deserializations?
You could argue that users could just conform their type to `Codable`, but not everything can easily be `Codable`. For example, `[String: String]`.

If we did this, it could make sense to make built-in types conform to this too (https://www.vadimbulavin.com/advanced-guide-to-userdefaults-in-swift/).

**I'm looking for feedback on this.**

There's another solution to this we could consider, and that would be to ship with [`AnyCodable`](https://github.com/Flight-School/AnyCodable) support.

hank121314:
## Summary

Fixes #52 .

This pr intent to let user do custom serializations and deserializations.
And there are some additional features.

## Features

1. Since UserDefaults support store array and dictionary natively. This pr add dictionary and array to isNativelySupportedType, so we can use array and dictionary natively.

2. Add DefaultsRawRepresentableBridge to store enum, no need to confirm to Codable
```swift
// Before 
enum FixtureEnum: String, Codable {
	case tenMinutes = "10 Minutes"
	case halfHour = "30 Minutes"
	case oneHour = "1 Hour"
}

// After
enum FixtureEnum: String, DefaultsSerializable {
	case tenMinutes = "10 Minutes"
	case halfHour = "30 Minutes"
	case oneHour = "1 Hour"
}
```

3. For those using custom  serializations and deserializations, we also support use the type in array or dictionary type.
```swift
enum FixtureIntEnum: Int, DefaultsSerializable {
	case tenMinutes = 10
	case halfHour = 30
	case oneHour = 60
}

let key = Defaults.Key<[FixtureEnum]>("array_enum", default: [.tenMinutes, .halfHour])
```

## Breaking Changes

But these are some breaking change: 
1. Before this pr, Value of Key need to confirm to protocol Codable. After this pr, Value of Key need to confirm to protocol DefaultsSerializable.
```swift
// Before
struct Unicorn: Codable {
	var isUnicorn: Bool
}

// After
struct Unicorn: Codable, DefaultsSerializable {
	var isUnicorn: Bool
}
```

Please feel free to critique and thanks for your code review!


Tests/DefaultsTests/DefaultsTests.swift
```diff
@@ -5,16 +5,27 @@ import XCTest
...
 let fixtureURL2 = URL(string: "https://example.com")!
 
-enum FixtureEnum: String, Codable {
+enum FixtureEnum: String, DefaultsSerializable {
 	case tenMinutes = "10 Minutes"
 	case halfHour = "30 Minutes"
 	case oneHour = "1 Hour"
 }
 
+enum FixtureIntEnum: Int, DefaultsSerializable {
```
sindresorhus, line 17: Can you also add a test for a enum like `enum FixtureIntEnum: Int, Codable, DefaultsSerializable {`?

Sources/Defaults/DefaultsBridge.swift
```diff
@@ -0,0 +1,185 @@
...
+	associatedtype Serializable
+
+	// Serialize Value to Serializable before we store it in UserDefaults
+	func serialize(_ value: Value?) -> Serializable?
+
+	// Deserialize Serializable to Value
+	func deserialize(_ object: Serializable?) -> Value?
+}
+
+public struct DefaultsURLBridge: DefaultsBridge {
```
sindresorhus, line 17: Can you nest these inside `extension Defaults {}`? So they're "kinda" namespaced.

sindresorhus:
Thanks for working on this 🙌🏻  This is already looking really good!

sindresorhus:
Some quick thoughts. I will review more thorough tomorrow.

- I would like to see even more tests. This is a pretty big change. I want to ensure it doesn't cause any unforseen breakage.
- Needs readme updates, updating the examples, and also a section on how to conform to an existing type.
- Does it support a dictionary in an array? `[[String: Foo]]`. If not, that would be useful. And how about array as dictionary value?
- Can you add a typealias for `typealias Serializable = DefaultsSerializable` inside `Defaults` so we can use `Defaults.Serializable` instead? That's the best way to have some kind of namespacing in Swift.
- Could all the extra handling of `NSSecureCoding` be simplified/unified by making it conform to `DefaultsSerializable`? Would be nice to simplify all that.
- I noticed `URL` is now saved as `Data`. It was previously saved as `String` (wrapped in `"`). Is there any reason it has to be saved as `Data`? Saving it as `String` makes it easier to debug.
- I fear that RawRepresentable enum are going to be a breaking change. Before you had `enum Foo: Int, Codable` and it was saved by encoding as JSON. Now it will encode as the raw value instead. Any thoughts on how we can handle that? Either we need some kind of automatic migration, or we need to have some kind of global switch `Defaults.storageVersion = .v2` or something.

sindresorhus (changes requested):

sindresorhus:
> Before this pr, Value of Key need to confirm to protocol Codable. After this pr, Value of Key need to confirm to protocol DefaultsSerializable.

This is too bad, but I think the new flexibility makes it totally worth it. Too bad we cannot do `extension Codable: DefaultsSerializable {}`...

Sources/Defaults/DefaultsBridge.swift
```diff
@@ -0,0 +1,185 @@
+import Foundation
+
+public protocol DefaultsBridge {
+	// The type of Value of Key<Value>
```
sindresorhus, line 4: Super nitpick, but end comments with dot. And:
```suggestion
	/// The type of `Value` of `Key<Value>`.
```

Sources/Defaults/UserDefaults.swift
```diff
@@ -1,22 +1,13 @@
...
-		else {
-			return nil
-		}
-
-		do {
-			return (try JSONDecoder().decode([Value].self, from: data)).first
-		} catch {
-			print(error)
+	private func _get<Value: DefaultsSerializable>(_ key: String) -> Value? {
+		if let any_object = object(forKey: key) {
```
sindresorhus, line 5:
```suggestion
		if let anyObject = object(forKey: key) {
```

sindresorhus:
This fixes https://github.com/sindresorhus/Defaults/issues/53 too, right?

Sources/Defaults/DefaultsBridge.swift
```diff
@@ -0,0 +1,185 @@
...
+	public func serialize(_ value: URL?) -> Any? {
+		if let value = value {
+			if #available(macOS 10.13, watchOS 4.0, macOSApplicationExtension 10.13, watchOSApplicationExtension 4.0, *) {
+				let data = try? NSKeyedArchiver.archivedData(withRootObject: value, requiringSecureCoding: false)
+				return data
+			}
+
+			// If system does not support NSKeyedArchiver, encode url to store in userDefaults
+			do {
+				// Some codable values like URL and enum are encoded as a top-level
```
sindresorhus, line 27: Are you sure this only applies to `URL` and enum?

Sources/Defaults/DefaultsBridge.swift
```diff
@@ -0,0 +1,185 @@
...
+	func serialize(_ value: Value?) -> Serializable?
+
+	// Deserialize Serializable to Value
+	func deserialize(_ object: Serializable?) -> Value?
+}
+
+public struct DefaultsURLBridge: DefaultsBridge {
+	public func serialize(_ value: URL?) -> Any? {
+		if let value = value {
+			if #available(macOS 10.13, watchOS 4.0, macOSApplicationExtension 10.13, watchOSApplicationExtension 4.0, *) {
```
sindresorhus, line 20: This is not correct. The way to encode URL should not depend on the platform version. That was not how the old logic worked.

Sources/Defaults/DefaultsBridge.swift
```diff
@@ -0,0 +1,185 @@
...
+	public typealias Serializable = [Bridge.Serializable]
+
+	private let bridge: Bridge
+
+	init(bridge: Bridge) {
+		self.bridge = bridge
+	}
+
+	public func serialize(_ value: Value?) -> Serializable? {
+		if let value = value as? [Value.Property] {
```
sindresorhus, line 147: Use `guard` whenever possible. Applies in many places.

Sources/Defaults/Utilities.swift
```diff
@@ -146,3 +144,58 @@ extension DispatchQueue {
...
+extension DefaultsSerializable where Self: Codable {
+	public static var bridge: DefaultsCodableBridge<Self> { return DefaultsCodableBridge<Self>() }
+}
+
+extension DefaultsSerializable where Self: RawRepresentable {
+	public static var bridge: DefaultsRawRepresentableBridge<Self> { return DefaultsRawRepresentableBridge() }
+}
+
+extension URL: DefaultsSerializable {
+	public static var bridge: DefaultsURLBridge { return DefaultsURLBridge() }
```
sindresorhus, line 192:
```suggestion
	public static let bridge = DefaultsURLBridge()
```

Applies to the others too.

Sources/Defaults/Utilities.swift
```diff
@@ -146,3 +144,58 @@ extension DispatchQueue {
 		}
 	}
 }
+
+public protocol NativelySupportedType {
```
sindresorhus, line 148: I think this and below should be in a separate file. It's not really utilities.

Sources/Defaults/Utilities.swift
```diff
@@ -5,15 +5,13 @@ extension Decodable {
...
 		}
-
 		self = value
 	}
 
 	init?(jsonString: String) {
 		guard let data = jsonString.data(using: .utf8) else {
 			return nil
 		}
-
```
sindresorhus, line 16: Don't do unrelated changes.

hank121314:
I have push a commit with the following changes:
1. Add protocol `DefaultsCodableBridge` and extension to handle all `Codable` Value.
2. Add typealias for `DefaultsSerializable`, `DefaultsBridge`, `DefaultsNativelySupportedType`, `DefaultsCodableBridge ` in `Defaults`.
3. Add `NSSecureCodingBridge` to simplify `NSSecureKey`.
4. Add `RawRepresentableCodableBridge` to handle `enum SomeEnum: Int, Codable, DefaultsSerializable`. Value which confirm protocol Codable and RawRepresentable will transform by the bridge.
5. Keep URL logic as the old ones(Codable).
6. Add `DefaultsArrayTests` to test array type.

And answer to the question:
> * I would like to see even more tests. This is a pretty big change. I want to ensure it doesn't cause any unforseen breakage.
- It is always good to have more tests. I will add more tests in the following commit. I have already added `DefaultsArrayTests` file. Can I add the others test file in this format?
> * Needs readme updates, updating the examples, and also a section on how to conform to an existing type.
- Readme update is required. But I consider to do it last of all.
> * I noticed `URL` is now saved as `Data`. It was previously saved as `String` (wrapped in `"`). Is there any reason it has to be saved as `Data`? Saving it as `String` makes it easier to debug.
- For the URL, at first I just wanna it to be like `UserDefaults.standard.url(forKey: name)`. So I reference some code from: 
[swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation/blob/99666897358504282fc9cc9d7dcea2add22b0d9a/Sources/Foundation/UserDefaults.swift#L229-L243). But as for debugging, I think we should keep url transformation as the old ones.
> * I fear that RawRepresentable enum are going to be a breaking change. Before you had `enum Foo: Int, Codable` and it was saved by encoding as JSON. Now it will encode as the raw value instead. Any thoughts on how we can handle that? Either we need some kind of automatic migration, or we need to have some kind of global switch `Defaults.storageVersion = .v2` or something.
- I think the `RawRepresentableCodableBridge` might be helpful to this situation? Or we still need some migration function to let Old Codable RawRepresentable to new RawRepresentable?

hank121314:
> This fixes #53 too, right?

Yes. I think it should. Maybe also #51 ?

sindresorhus:
> Can I add the others test file in this format?

Yes, would be great if you could split them up into more files.

> I think the RawRepresentableCodableBridge might be helpful to this situation? Or we still need some migration function to let old Codable RawRepresentable to be new RawValue RawRepresentable?

So what you're saying is that if the user keeps it `enum Foo: Int, Codable, Defaults.Serializable` it will use the old compatible format, but if they remove the `Codable` (`enum Foo: Int, Defaults.Serializable`) it will switch to the new format? If so, that is a perfect solution and gives the user the choice to migrate stuff themselves or keep it as is.

Sources/Defaults/SwiftUI.swift
```diff
@@ -4,7 +4,7 @@ import Combine
 
 @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 extension Defaults {
-	final class Observable<Value: Codable>: ObservableObject {
+	final class Observable<Value: DefaultsSerializable>: ObservableObject {
```
sindresorhus, line 7:
```suggestion
	final class Observable<Value: Serializable>: ObservableObject {
```

Probably in other places too.

Sources/Defaults/UserDefaults.swift
```diff
@@ -127,6 +54,12 @@ extension UserDefaults {
 			is Data.Type,
 			is Data?.Type:
 			return true
```
sindresorhus, line 56: Do we still need this now that we have the bridges? Couldn't we just rely on the bridges to do the correct thing?

For example, if we defined something like this:

```swift
extension Float: Defaults.NativelySupportedType {}
```

sindresorhus (changes requested):

Sources/Defaults/UserDefaults.swift
hank121314: I think the bridges should use to compile time check, and `isNativelySupportedType` is for runtime check?

sindresorhus: I'm not sure that is the case as most of this can be determined at compile time. Would be worth looking into at least.

For example, https://www.vadimbulavin.com/advanced-guide-to-userdefaults-in-swift/ is only using the protocol constraints.

hank121314: Yes. But in this project we still need `isNativelySupportedType` to specify  if the `Value` need to be serialized.
If we remove `isNativelySupportedType` only using the protocol constraints. we need to create bridge for all `Defaults.Serializable` like [SwiftyUserDefaults/DefaultsBridges.swift](https://github.com/sunshinejr/SwiftyUserDefaults/blob/master/Sources/DefaultsBridges.swift)

sindresorhus: Ah, right. Good point.

hank121314:
> So what you're saying is that if the user keeps it `enum Foo: Int, Codable, Defaults.Serializable` it will use the old compatible format, but if they remove the `Codable` (`enum Foo: Int, Defaults.Serializable`) it will switch to the new format? If so, that is a perfect solution and gives the user the choice to migrate stuff themselves or keep it as is.

Yes, but after user switching to new format, the old format value might not be available.

And few points about using protocol constraints instead of `isNativelySupportedType `.

I divided `Defaults.Serializable` into `Defaults.NativelySupportedType` and `Defaults.Serializable` which are two independent protocol.

But when I trying to separate the `@Default` for SwiftUI.
I encountered some problems about the `@propertyWrapper`. It cannot constrained extension. 😢  
It has been reported to [bugs.swift](https://bugs.swift.org/browse/SR-12504?jql=labels%20%3D%20PropertyWrappers)

So I use some tricky way trying to implement both Defaults.NativelySupportedType and Defaults.Serializable in `@Default`.
In [Defaults/SwiftUI.swift](https://github.com/hank121314/Defaults/blob/0307982681123fdd502673e748af916fb1a9cf47/Sources/Defaults/SwiftUI.swift#L149-L171).

If you think that tricky way is acceptable I will merge it into this pr.

Thanks for your assistance!

sindresorhus:
> Yes, but after user switching to new format, the old format value might not be available.

I'm not sure exactly what you mean by this?

> I divided Defaults.Serializable into Defaults.NativelySupportedType and Defaults.Serializable which are two independent protocol.

What's the benefit of doing this?

> If you think that tricky way is acceptable I will merge it into this pr.
Or is there any better way of doing it?

I quickly looked at it and I cannot think of a better way to do it either. At least add some code comments about why it's like this, link to the Swift bug report and say how it can be simplified when the Swift issue is fixed.

Sources/Defaults/Defaults+Extensions.swift
```diff
@@ -0,0 +1,63 @@
...
+@available(iOS 11.0, macOS 10.13, tvOS 11.0, watchOS 4.0, iOSApplicationExtension 11.0, macOSApplicationExtension 10.13, tvOSApplicationExtension 11.0, watchOSApplicationExtension 4.0, *)
+extension Defaults.Serializable where Self: NSSecureCoding {
+	public static var bridge: Defaults.NSSecureCodingBridge<Self> { return Defaults.NSSecureCodingBridge() }
+}
+
+extension Optional: Defaults.Serializable where Wrapped: Defaults.Serializable {
+	public static var bridge: Defaults.OptionalBridge<Wrapped.Bridge> { return Defaults.OptionalBridge(bridge: Wrapped.bridge) }
+}
+
+extension Array: Defaults.Serializable where Element: Defaults.Serializable {
```
sindresorhus, line 45: Can you add support for `Set` or better yet `SetAlgebra` too?

sindresorhus:
Is there a way we could support any `Collection`?

sindresorhus:
Could you add built-in support for `NSColor`/`UIColor` too? They're often needed to be saved and they're securecoding compatible, so I think they could just be `extension NSColor: Defaults.Serializable {}`, but needs tests.

sindresorhus (changes requested):

sindresorhus:
Is there any way we can make this work without it having to be annotated as `Codable`?

```swift
struct Foo: Defaults.Serializable {
	let a: String
	let b: Int
}
```

sindresorhus:
I discovered another breaking change.

```
extension Defaults.Keys {
	static let foo = Key<[String]>("foo", default: [])
}

Defaults[.foo] = ["a", "b", "c"]
```

Results in a native plist array of strings.

Previously, it would result in a native plist string with `["a","b","c"]`.

We need to add a way for the user to get back the previous behavior and document it. Maybe force the user to wrap the array in a wrapper that ensures the array is encoded/decoded as a codable like before?

hank121314:
> I'm not sure exactly what you mean by this? 

If user use `enum SomeEnum: Int, Codable` previously, 
they have to use  `enum SomeEnum: Int, Codable, Defaults.Serializable` now.
If they remove `Codable` from `SomeEnum`, the old value in UserDefaults cannot be serialized, since the old value was codable.

> What's the benefit of doing this?

We can remove `isNativelySupportedType` and only use protocol constrains.

hank121314:
> Could you add built-in support for `NSColor`/`UIColor` too? They're often needed to be saved and they're securecoding compatible, so I think they could just be `extension NSColor: Defaults.Serializable {}`, but needs tests.

Should we use NSSecureCodingBridge directly?
I think we might be able to serialize it to hexString?
It is easier to debug.
Like this:
```swift
public struct NSColorBridge: Defaults.Bridge {
    public func serialize(_ value: String?) -> NSColor? {
        guard let value = value,
                    let hex = Int(value, radix: 16)
        else {
            return nil
        }

        let red = UInt(hex >> 24)
        let green = UInt((hex & 0x00FF0000) >> 16)
        let blue = UInt((hex & 0x0000FF00) >> 8)
        let alpha = UInt(hex & 0xFF)

        return NSColor(calibratedRed: CGFloat(red) / CGFloat(0xFF), green: CGFloat(green) / CGFloat(0xFF), blue: CGFloat(blue) / CGFloat(0xFF), alpha: CGFloat(alpha) / CGFloat(0xFF))
    }

    public func deserialize(_ object: NSColor?) -> String? {
        guard let object = object else {
            return nil
        }
        let red = object.redComponent * 0xFF
        let green = object.greenComponent * 0xFF
        let blue = object.blueComponent * 0xFF
        let alpha = object.alphaComponent * 0xFF

        let hex = String(format: "%02X%02X%02X%02X", red, green, blue, alpha)

        return hex
    }
}
```

hank121314:
> I discovered another breaking change.
> 
> ```swift
> extension Defaults.Keys {
> 	static let foo = Key<[String]>("foo", default: [])
> }
> 
> Defaults[.foo] = ["a", "b", "c"]
> ```
> 
> Results in a native plist array of strings.
> 
> Previously, it would result in a native plist string with `["a","b","c"]`.
> 
> We need to add a way for the user to get back the previous behavior and document it. Maybe force the user to wrap the array in a wrapper that ensures the array is encoded/decoded as a codable like before? This will also need tests.

I have few thoughts about this changes.
Let `Defaults.NativelySupportedType` confirm to `Codable`.
And use JSONDecoder to decode it when `object(forKey: String)` can not be casted to `Value`.
After decoded, call `_set(_ key: String, to value: Value)` to overwrite the old codable value.

More details in: [Defaults/UserDefaults.swift](https://github.com/hank121314/Defaults/blob/4511f718e424c124a4cb09d3466e9ae59472f223/Sources/Defaults/UserDefaults.swift#L4-L16)

And few tests which are passed:
[DefaultsArrayTests.swift](https://github.com/hank121314/Defaults/blob/4511f718e424c124a4cb09d3466e9ae59472f223/Tests/DefaultsTests/DefaultsArrayTests.swift#L193-L201)
[DefaultsDictionaryTests.swift](https://github.com/hank121314/Defaults/blob/4511f718e424c124a4cb09d3466e9ae59472f223/Tests/DefaultsTests/DefaultsDictionaryTests.swift#L185-L191)

sindresorhus:
> Should we use NSSecureCodingBridge directly?

Yes

> I think we might be able to serialize it to hexString?

While it's easier to debug, it also might lose information as NSColor also contains a color profile. I think it's safer to archive it.

hank121314:
> Is there any way we can make this work without it having to be annotated as `Codable`?
> 
> ```swift
> struct Foo: Defaults.Serializable {
> 	let a: String
> 	let b: Int
> }
> ```

Yes, but it should have static property` bridge` which protocol conform to `Defaults.Bridge`.
For more example: [Tests/DefaultsCustomBridgeTests.swift](https://github.com/hank121314/Defaults/blob/develop/Tests/DefaultsTests/DefaultsCustomBridgeTests.swift)

Sources/Defaults/Defaults+Extensions.swift
```diff
@@ -17,6 +16,10 @@ extension UInt32: Defaults.NativelySupportedType {}
 extension Int64: Defaults.NativelySupportedType {}
 extension UInt64: Defaults.NativelySupportedType {}
 
+#if os(macOS)
+extension CGFloat: Defaults.NativelySupportedType{}
+#endif
```
sindresorhus: Why macOS only?

sindresorhus:
> We can remove isNativelySupportedType and only use protocol constrains.

👍

Sources/Defaults/Defaults+Extensions.swift
hank121314: Sorry, it was a typo. 😭

hank121314:
In order to support `Set`.
I have to create a new protocol name `Defaults.GenericCollectionType`.
There are some duplicate code because swift cannot use multiple conformances. 😢

hank121314:
I've thought about the protocol constraint in the past few days.
I think we should add `isNativelySupportedType` back.
In our codebase, there are so many workaround trying to resolve the restriction of swift.
I add `isNativelySupportedType` back as a protocol static property.
So we can check whether it can store natively or not.
```
extension String: Defaults.Serializable {
	public static let isNativelySupportedType = true
}
```

Thanks for your code review!

hank121314:
Support any `Collection`.
Class which conform protocol `Collection` and `Defaults.Serializable` should also conform `ExpressibleByArrayLiteral`.
Because we need `init(arrayLiteral: Self.ArrayLiteralElement...)` to create the instance in function `bridge.deserialize`.

hank121314:
Support any `SetAlgebra`.
I also create a protocol `Defaults. SetAlgebraSerializable`.
Because we still need the initializer to create the instance 😄 .
For the example: 
- Native supported type: [Defaults/Tests/DefaultsSetAlgebraTests.swift](https://github.com/hank121314/Defaults/blob/develop/Tests/DefaultsTests/DefaultsSetAlgebraTests.swift)
- User Customization type: [Defaults/Tests/DefaultsSetAlgebraCustomElementTests.swift](https://github.com/hank121314/Defaults/blob/develop/Tests/DefaultsTests/DefaultsSetAlgebraCustomElementTests.swift)

Sources/Defaults/Defaults+Extensions.swift
hank121314: Added! 😄

sindresorhus:
Did you implement what you discussed in https://github.com/sindresorhus/Defaults/pull/57#issuecomment-751616020 ?

I just tried the latest commit with my Dato app and I noticed that this:

```swift
extension Defaults.Keys {
	static let enabledCalendars = Key<Set<String>>("enabledCalendars", default: [])
}
```

which had entries would not be read anymore and defaulted to empty.

*(Also add tests for this)*

sindresorhus:
With:

```swift
struct TimeZone: Codable, Defaults.Serializable {
	let id: String
	let name: String
}

extension Defaults.Keys {
	static let timeZonesInMenu = Key<[TimeZone]>("timeZonesInMenu", default: [])
}
```

and with the current Defaults version it produces a plist `String` with the contents:

```
[{"id":"0D4EB864-DDAF-4A39-9AAF-01D0C31581E7","name":"Foo"}]
```

with the latest PR commit, it produces a plist Array of String's where each item contains:

```
{"id":"0D4EB864-DDAF-4A39-9AAF-01D0C31581E7","name":"Foo"}
```

Would it be possible to encode the array element as a plist dictionary instead of a JSON string?

Sources/Defaults/Defaults+Extensions.swift
```diff
@@ -0,0 +1,113 @@
...
+	public static var isNativelySupportedType: Bool { Wrapped.isNativelySupportedType }
+	public static var bridge: Defaults.OptionalBridge<Wrapped> { Defaults.OptionalBridge() }
+}
+
+extension Array: Defaults.Serializable where Element: Defaults.Serializable {
+	public static var isNativelySupportedType: Bool { Element.isNativelySupportedType }
+	public static var bridge: Defaults.ArrayBridge<Element> { Defaults.ArrayBridge() }
+}
+
+extension Dictionary: Defaults.Serializable where Key == String, Value: Defaults.Serializable {
```
sindresorhus, line 104:
```suggestion
extension Dictionary: Defaults.Serializable where Key: ExpressibleByStringLiteral, Value: Defaults.Serializable {
```

?

Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -0,0 +1,258 @@
...
+		guard let jsonString = object else {
+			return nil
+		}
+
+		return [Value].init(jsonString: "[\(jsonString)]")?.first
+	}
+}
+
+extension Defaults {
+	public struct TopLevelCodableBridge<Value: Codable>: CodableBridge {}
```
sindresorhus, line 36: Needs a doc comment.

sindresorhus: Can you split out all of these into separate extensions instead. To reduce nesting.

```swift
extension Defaults {
	public struct TopLevelCodableBridge<Value: Codable>: CodableBridge {}
}
```

sindresorhus (changes requested):

sindresorhus:
Why do you need both a `SetBridge` and `SetAlgebraBridge`? `Set` conforms to `SetAlgebra`.

hank121314:
> Did you implement what you discussed in [#57 (comment)](https://github.com/sindresorhus/Defaults/pull/57#issuecomment-751616020) ?
> 
> I just tried the latest commit with my Dato app and I noticed that this:
> 
> ```swift
> extension Defaults.Keys {
> 	static let enabledCalendars = Key<Set<String>>("enabledCalendars", default: [])
> }
> ```
> 
> which had entries would not be read anymore and defaulted to empty.
> 
> _(Also add tests for this)_

Oops! I only add the migration function to dictionary and array.
The latest commit should support the migration from codable array to `Set`.

Sources/Defaults/Defaults+Extensions.swift
hank121314: In [swift-corelibs-foundation](https://github.com/apple/swift-corelibs-foundation/blob/ef6f96ee82ea0f54252071c0ecadf5f01be9aecc/Foundation/UserDefaults.swift#L41), it seems that `Key` must be a String type?

hank121314:
> Why do you need both a `SetBridge` and `SetAlgebraBridge`? `Set` conforms to `SetAlgebra`.

Because `Set` conforms to `Sequence` but `SetAlgebra` not.
`Set` conforms to `Sequence`, so we can convert it to an array by `Array.init<S>(S)` and store it in the `UserDefaults`.
But `SetAlgebra` does not, so it is hard to convert it to an array.
Thats why we need `Defaults.SetAlgebraSerializable` protocol to convert it to an array.

sindresorhus:
> > Why do you need both a `SetBridge` and `SetAlgebraBridge`? `Set` conforms to `SetAlgebra`.
> 
> 
> 
> Because `Set` conforms to `Sequence` but `SetAlgebra` not.
> 
> `Set` conforms to `Sequence`, so we can convert it to an array with `Array.init<S>(S)` and store it in the `UserDefaults`.
> 
> But `SetAlgebra` does not, so it is hard to convert it to an array.
> 
> Thats why we need [`Defaults.SetAlgebraSerializable`](https://github.com/sindresorhus/Defaults/pull/57#issuecomment-755231393) protocol to convert it to an array.

Makes sense. Can you put the above text in a code comment? In case anyone reading the code has the same question.

hank121314:
There are two requirement for creating a bridge:

1. We can convert it to `Serializable` which is `UserDefaults` native supported type.
2. We can initialize `Value` with the `Serializable`.

For example:
1. We can convert `Set` into `Array` with `Array.init<S>(S)`.
2. We can initialize `Set` with `Array` using `Set.init<S>(S)`.

hank121314:
I am trying to add a `migration` function which can migrate codable json string into native support type.
For example: 
```swift
let text = "[\"a\",\"b\",\"c\"]"
let keyName = "arrayToNativeSet"
UserDefaults.standard.set(text, forKey: keyName)
let key = Defaults.Key<Set<String>?>(keyName)
key.migration()
```
After calling `key.migration()` the value in `UserDefaults` will be converted into a native array.

If you want to convert codable into the user customized type, please do following things.

1. Create struct name `TimeZone` which protocol conform to `Defaults.Serializable`. It should have a static property bridge which have `serialize` and `deserialize` function (protocol conform to `Defaults.Bridge`).

2. Create a codable struct name `CodableTimeZone` which protocol conform to `Defaults.CodableType`. It should have a toNative function which will return `TimeZone`

3. Let `TimeZone` protocol conform to `Defaults.NativeType`, associated type `CodableForm` to `CodableTimeZone`

```swift
struct TimeZone: Defaults.Serializable, Defaults.NativeType, Hashable {
	/// Associated `CodableForm` to `CodableTimeZone`
	typealias CodableForm = CodableTimeZone

	var id: String
	var name: String

	static let bridge = TimeZoneBridge()
}

struct CodableTimeZone: Defaults.Serializable, Defaults.CodableType {
	var id: String
	var name: String

	/// Convert from `Codable` to `Native`
	func toNative() -> TimeZone {
		TimeZone(id: id, name: name)
	}
}

struct TimeZoneBridge: Defaults.Bridge {
	typealias Value = TimeZone
	typealias Serializable = [String: Any]

	func serialize(_ value: TimeZone?) -> Serializable? {
		guard let value = value else {
			return nil
		}

		return ["id": value.id, "name": value.name]
	}

	func deserialize(_ object: Serializable?) -> TimeZone? {
		guard
			let dictionary = object,
			let id = dictionary["id"] as? String,
			let name = dictionary["name"] as? String
		else {
			return nil
		}

		return TimeZone(id: id, name: name)
	}
}
```

4. After these instructions we can use it like this.
```swift
func testArrayAndCodableElementToNativeArray() {
  let text = "[{\"id\":\"0\", \"name\": \"Asia/Taipei\"}]"
  let keyName = "codableArrayAndCodableElementToNativeArray"
  UserDefaults.standard.set(text, forKey: keyName)
  let key = Defaults.Key<[TimeZone]?>(keyName)
  key.migration()
  XCTAssertEqual(Defaults[key]?[0].id, "0")
  let newName = "Asia/Tokyo"
  Defaults[key]?[0].name = newName
  XCTAssertEqual(Defaults[key]?[0].name, newName
}
```

Sources/Defaults/Defaults+Extensions.swift
hank121314: I think using `LosslessStringConvertible & Hashable` is more comprehensive.
We can initialize dictionary's `Key` by `String` with `init?(_ description: String)`
And initialize `String` by dictionary's `Key` with `init<T>(_ value: T) where T : LosslessStringConvertible`

sindresorhus:
With the latest changes, I need to manually migrate using `Defaults.migrate`, correct? There's no auto-migration?

I tried, but I had a type that was previously `struct ChosenTimeZone: Codable, Defaults.Serializable {}`, but the migration function complains:

> Referencing static method 'migration' on 'Array' requires that 'ChosenTimeZone' conform to 'DefaultsNativeType'

I change it to `struct ChosenTimeZone: Codable, Defaults.NativeType {}`, but it's not immediate clear what I should put in `typealias CodableForm`. I want to keep the type `Codable`. All I want is for it to continue working (it's not being read without migration), and to use the native array type. The default key is `static let timeZonesInMenu = Key<[ChosenTimeZone]>("timeZonesInMenu", default: [])`.

sindresorhus:
Do we really need to expose the two types `Defaults.Serializable` and `Defaults.NativeType>` to the end-user? Would be nice if all they needed to know was some kind of `typealias` for `Defaults.Serializable & Defaults.NativeType>`.

sindresorhus:
> If you want to convert codable into the user customized type, please do following things.

This feels a bit overdesigned. Can we simplify it? Do we really need to enable this ability?

sindresorhus:
I would really like the migration thing to be a one-off kinda thing that doesn't complicate the types and usage of `Defaults`. I feel the migration stuff should be so decoupled from `Defaults` that sometime in the future we could remove the migration code and everything should continue working without much changes.

Sources/Defaults/Migration.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+extension Defaults {
+	public static func migration<Value: Defaults.Serializable & Defaults.NativeType>(_ keys: Key<Value>...) {
```
sindresorhus, line 4:
```suggestion
	public static func migrate<Value: Defaults.Serializable & Defaults.NativeType>(_ keys: Key<Value>...) {
```

I think the function name should be more specific to what's being migrated. We might add other kinds of migrations in the future.

Sources/Defaults/SwiftUI.swift
```diff
@@ -40,7 +40,7 @@ extension Defaults {
 
 @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 @propertyWrapper
-public struct Default<Value: Codable>: DynamicProperty {
+public struct Default<Value: Defaults.Serializable>: DynamicProperty {
```
sindresorhus, line 43:
```suggestion
public struct Default<Value: Serializable>: DynamicProperty {
```

hank121314:
> With the latest changes, I need to manually migrate using `Defaults.migrate`, correct? There's no auto-migration?
> 
> I tried, but I had a type that was previously `struct ChosenTimeZone: Codable, Defaults.Serializable {}`, but the migration function complains:
> 
> > Referencing static method 'migration' on 'Array' requires that 'ChosenTimeZone' conform to 'DefaultsNativeType'
> 
> I change it to `struct ChosenTimeZone: Codable, Defaults.NativeType {}`, but it's not immediate clear what I should put in `typealias CodableForm`. I want to keep the type `Codable`. All I want is for it to continue working (it's not being read without migration), and to use the native array type. The default key is `static let timeZonesInMenu = Key<[ChosenTimeZone]>("timeZonesInMenu", default: [])`.

I see.
So what you want to do is storing `Key<[ChosenTimeZone]>` as a native array and it's element is a plist dictionary?
Since `ChosenTimeZone` conform to `Codable`, it will goes to `TopLevelCodableBridge` by default.
If you need to convert a `Codable` to dictionary, it will need a custom `bridge` to do the `serialize` and the `deserialize`.
`ChosenTimeZone` conform to `Codable` that means the `CodableForm` is `Self`.
Full example: 
```swift
private struct ChosenTimeZone: Defaults.NativeType, Defaults.CodableType {
	typealias CodableForm = Self

	var id: String
	var name: String

	static let bridge = ChosenTimeZoneBridge()

	func toNative() -> Self {
		self
	}
}

private struct ChosenTimeZoneBridge: Defaults.Bridge {
	typealias Value = ChosenTimeZone
	typealias Serializable = [String: Any]

	func serialize(_ value: Value?) -> Serializable? {
		guard let value = value else {
			return nil
		}

		return ["id": value.id, "name": value.name]
	}

	func deserialize(_ object: Serializable?) -> Value? {
		guard
			let dictionary = object,
			let id = dictionary["id"] as? String,
			let name = dictionary["name"] as? String
		else {
			return nil
		}

		return ChosenTimeZone(id: id, name: name)
	}
}

let keyName = "codableToNativeCodableArrayType"
UserDefaults.standard.set("[{\"id\":\"0\",\"name\":\"Asia\\/Taipei\"}]", forKey: keyName)
let key = Defaults.Key<[ChosenTimeZone]?>(keyName)
Defaults.migration(key)
XCTAssertEqual(Defaults[key]?[0].id, "0")
let newName = "Asia/Tokyo"
Defaults[key]?[0].name = newName
XCTAssertEqual(Defaults[key]?[0].name, newName)
```

hank121314:
> Do we really need to expose the two types `Defaults.Serializable` and `Defaults.NativeType>` to the end-user? Would be nice if all they needed to know was some kind of `typealias` for `Defaults.Serializable & Defaults.NativeType>`.

I think `Defaults.NativeType` should conform to `Defaults.Serializable` this will reduce some composition.

sindresorhus:
> So what you want to do is storing Key<[ChosenTimeZone]> as a native array and it's element is a plist dictionary?

Maybe at some point, but right now I just want to be able to easily migrate it to the new Defaults version, and ideally have it use native array (with Codable elements, not dictionary) automatically.

> Since ChosenTimeZone conform to Codable, it will goes to TopLevelCodableBridge by default.

We should use the native types whenever possible, automatically. In this case, it should use native array for the top-level container and Codable for the array elements.

hank121314:
> Maybe at some point, but right now I just want to be able to easily migrate it to the new Defaults version, and ideally have it use native array (with Codable elements, not dictionary) automatically.

If you want it to be codable elements, bridge will no longer be needed.
Change `ChosenTimeZone` like this.
```swift
private struct ChosenTimeZone: Defaults.NativeType, Defaults.CodableType {
	typealias CodableForm = Self

	var id: String
	var name: String

	func toNative() -> Self {
		self
	}
}
```

> We should use the native types whenever possible, automatically. In this case, it should use native array for the top-level container and Codable for the array elements.

Sorry for the confusion.
`ChosenTimeZone` will goes to `TopLevelCodableBridge` by default. But `[ChosenTimeZone]` will goes to `ArrayBridge` and then goes to `TopLevelCodableBridge`. So it will be a native array for the top-level container and Codable for the array elements.

hank121314:
> > If you want to convert codable into the user customized type, please do following things.
> 
> This feels a bit overdesigned. Can we simplify it? Do we really need to enable this ability?

The only way to simplify it is using like https://github.com/sindresorhus/Defaults/pull/57#issuecomment-781107294.
I think this ability might not be necessary, but there might have some user who want to store it as their customize type rather than `Codable`.

hank121314:
> I would really like the migration thing to be a one-off kinda thing that doesn't complicate the types and usage of `Defaults`. I feel the migration stuff should be so decoupled from `Defaults` that sometime in the future we could remove the migration code and everything should continue working without much changes.

In the latest commit, I put all migration related code to `Migration` Group. 
When we have to remove the migration code, just remove the `Migration` Group and everything should work well. 
Is this an acceptable solutions?

Sources/Defaults/Migration.swift
hank121314: How about `migrateFromCodableToNative`?

Sources/Defaults/SwiftUI.swift
hank121314: `@Default` is not in `extension Defaults` this might be typo?

sindresorhus:
> I think this ability might not be necessary, but there might have some user who want to store it as their customize type rather than Codable.

I'm fine with having extensibility, but I think the default behavior should be simple without much boilerplate. Like it currently is.

I like how I can just make something codable in the current version and it **just works** with Defaults:

```swift
struct ChosenTimeZone: Codable {
	var id: String
	var name: String
}
```

sindresorhus:
Is there any way we could allow this:

```swift
struct ChosenTimeZone: Codable, Defaults.Serializable {
	var id: String
	var name: String
}
```

without any of the boilerplate stuff and have `Key<[ChosenTimeZone]>` serialized to a native array with codable elements?

And if the user wanted it to be stored as a dictionary instead of a serialized codable, they could opt into the more boilerplaty setup that you illustrated above with some bridging.

hank121314:
> Is there any way we could allow this:
> 
> ```swift
> struct ChosenTimeZone: Codable, Defaults.Serializable {
> 	var id: String
> 	var name: String
> }
> ```
> 
> without any of the boilerplate stuff and have `Key<[ChosenTimeZone]>` serialized to a native array with codable elements?

Yes, this kind of struct is available if there is no need to do the migration.
See the test for details.
[DefaultsCodableTests.swift](https://github.com/hank121314/Defaults/blob/develop/Tests/DefaultsTests/DefaultsCodableTests.swift)

> And if the user wanted it to be stored as a dictionary instead of a serialized codable, they could opt into the more boilerplaty setup that you illustrated above with some bridging.

That's right !

If there is a need to migrate UserDefaults value from json string to a native array with codable elements.
The struct should be like this in the latest commit.
```swift
private struct ChosenTimeZone: Defaults.NativeType, Defaults.CodableType {
	typealias CodableForm = Self

	var id: String
	var name: String

	func toNative() -> Self {
		self
	}
}
```
But I think there might have some more elegant way to do it.
I will investigate further

hank121314:
With the latest changes, we can migrate `Codable` element with `Defaults.migration` directly without any configuration.

For example:
```swift
extension Defaults.Keys {
	static let timezone = Defaults.Key<[ChosenTimeZone]?>("timezone")
}
private struct ChosenTimeZone: Defaults.Serializable & Codable {
	var id: String
	var name: String
}

Defaults.migration(timezone)
```
And `Defaults[.timezone]` should be readable.

I also create [migration.md](https://github.com/hank121314/Defaults/blob/develop/migration.md) to show how to do the others kind of migration.

Thanks for the code review 😄 !

sindresorhus:
> I also create migration.md to show how to do the others kind of migration.

Nice! Maybe you could also mention how it handles enum's? And also dictionary in array `Defaults.Key<[[String: Int]]?>("dictionaryStringIntInArray")`?

hank121314:
> > I also create migration.md to show how to do the others kind of migration.
> 
> Nice! Maybe you could also mention how it handles enum's? And also dictionary in array `Defaults.Key<[[String: Int]]?>("dictionaryStringIntInArray")`?

Added! 
It also mention how to migrate from `Codable enum` to `RawRepresentable`.

migration.md
```diff
@@ -1,58 +1,191 @@
...
 #### Migration steps
-1. Let `TimeZone` protocol conform to `Defaults.Serializable`
+1. Let `TimeZone` and `Period` protocol conform to `Defaults.Serializable`
 ```swift
 private struct TimeZone: Defaults.Serializable & Codable {
 	var id: String
 	var name: String
 }
+
+private enum Period: String, Defaults.Serializable & Codable {
```
sindresorhus, line 117: You use `Defaults.Serializable & Codable` here while you use `Defaults.Serializable, Codable` in other places. It should be consistent.

Sources/Defaults/SwiftUI.swift
sindresorhus: Yeah. Never mind.

migration.md
```diff
@@ -0,0 +1,267 @@
+## Migration Guide From v4.X to v5.X
```
sindresorhus, line 1: I think the migration guide should be split into what users **have to do** and what is optional.

migration.md
```diff
@@ -0,0 +1,267 @@
...
+
+### From `Codable struct` to `Codable struct`
```
sindresorhus, line 3:
```suggestion
### From `Codable` struct in Defaults v4 to `Codable` struct in Defaults v5
```

Same in other places.

migration.md
sindresorhus: The migration guide should be linked to from the top of the readme.

migration.md
```diff
@@ -0,0 +1,267 @@
...
+
+Before V4.X, `struct` have to conform to protocol `Codable` to store it as a json string.
```
sindresorhus, line 5:
```suggestion
Before v4, `struct` have to conform to protocol `Codable` to store it as a JSON string.  
```

migration.md
sindresorhus: For the two first optional migrations, we should explicitly point out that running `Defaults.migrate` is not needed.

sindresorhus (changes requested):

sindresorhus:
I'm very happy with how the code and API are now. You have been doing amazing work on this. 👍🏻

sindresorhus:
Would you be able to update the readme to reflect the new API? Don't worry about typos, etc. I will go through all docs later on to clean it up.

sindresorhus:
Any idea why the tests are failing?

migration.md
```diff
@@ -0,0 +1,267 @@
...
+	case tenMinutes = "10 Minutes"
+	case halfHour = "30 Minutes"
+	case oneHour = "1 Hour"
+}
+```
+2. Now `Defaults[.period]` should be readable.
+
+---
+
+### From `Codable Array/Dictionary` to `Native Array/Dictionary/Set`(With Native Supported Elements)
```
sindresorhus, line 64: It's a bit confusing that `Set` is not part of the "from" part.

migration.md
```diff
@@ -0,0 +1,267 @@
...
+Before v4.X, `Defaults` will store array/dictionary as a json string(`["a", "b", "c"]`).  
+
+After v5.X, `Defaults` will store it as a native array/dictionary with native supported elements(`[a, b, c]` ). 
+
+#### Before migration, your code should be like this
+```swift
+extension Defaults.Keys {
+	static let arrayString = Defaults.Key<[String]?>("arrayString")
+	static let dictionaryStringInt = Defaults.Key<[String: Int]?>("dictionaryStringInt")
+	static let dictionaryStringIntInArray = Defaults.Key<[[String: Int]]?>("dictionaryStringIntInArray")
```
sindresorhus, line 75: Can you add an entry using `Set` here too?

sindresorhus: Do we have tests for deep nesting? For example:

```swift
static let foo = Defaults.Key<[[String: [[String: Int]]]]?>("foo")
```

sindresorhus:
There are no breaking changes anymore, correct?

sindresorhus:
Hmm, actually, https://github.com/hank121314/Defaults/blob/develop/migration.md#from-codable-arraydictionary-to-native-arraydictionarysetwith-native-supported-elements is a breaking change. I think that should be more clearly pointed out by the migration guide. I think the migration guide needs a bullet list and an intro section that shows examples of what requires migration. It's a bit confusing at the moment which patterns *just work*, which *require* migration, and which where the migration is optional.

hank121314:
> I'm very happy with how the code and API are now. You have been doing amazing work on this. 👍🏻

I owe it all to you. Thank you so much for the good code review!
And also thanks for your effort to make the community better 😃 .

hank121314:
> Would you be able to update the readme to reflect the new API? Don't worry about typos, etc. I will go through all docs later on to clean it up.

Sure! Also I will try to improve my english ability in future.

hank121314:
> Any idea why the tests are failing?

Tests are failing cause by `swift test`, since macOS does not contain `UIKit`.
I add `#if !os(macOS)/#endif` at `DefaultsUIColorTests.swift`.
and add some action to test target `iOS` and `tvOS`.

migration.md
hank121314: > Can you add an entry using Set here too?

> Do we have tests for deep nesting? For example:

Added!

migration.md
hank121314: Sorry, I forgot that `Set` is `Codable` 😭 .

hank121314:
> Hmm, actually, [https://github.com/hank121314/Defaults/blob/develop/migration.md#from-codable-arraydictionary-to-native-arraydictionarysetwith-native-supported-elements](https://github.com/hank121314/Defaults/blob/develop/migration.md?rgh-link-date=2021-03-17T12%3A22%3A13Z#from-codable-arraydictionary-to-native-arraydictionarysetwith-native-supported-elements) is a breaking change. I think that should be more clearly pointed out by the migration guide. I think the migration guide needs a bullet list and an intro section that shows examples of what requires migration. It's a bit confusing at the moment which patterns _just work_, which _require_ migration, and which where the migration is optional.

Yes, this is the most critical breaking change in this pr.
I add a more precise migration guide at `migration.md`.
It mentions each issue that the user will encounter when migrating and demonstrate how to resolve it.

hank121314:
When I trying to edit the readme, I found one thing is worth mentioning.
[How can I store a dictionary of arbitrary values?](https://github.com/sindresorhus/Defaults#how-can-i-store-a-dictionary-of-arbitrary-values)
Below the section, Using `AnyCodable` framework to do the trick like this.
```swift
Defaults[.magic]["unicorn"] = "🦄"
Defaults[.magic]["number"] = 3
Defaults[.magic]["boolean"] = true
```
I wonder if there is a need to create an `AnySerializable` struct to replace `AnyCodable` with it?

sindresorhus:
> I wonder if there is a need to create an AnySerializable struct to replace AnyCodable with it?

Oh, yeah. Good idea! I think it should be a separate PR after this one though (that being, if you want to work on it), since this PR is already huge.

hank121314:
> > I wonder if there is a need to create an AnySerializable struct to replace AnyCodable with it?
> 
> Oh, yeah. Good idea! I think it should be a separate PR after this one though (that being, if you want to work on it), since this PR is already huge.

I am already working on it!
Will submit another PR after this PR get merged.

hank121314:
I change migration function from `Defaults.migration(keys...)` to `Defaults.migration(keys..., to: Version)`.
For example.
```swift
public enum Version: Int {
	case v5 = 5
}

Defaults.migration(.array, to: .v5)
```
If we have others migration in future, we can just specific a new version.

hank121314:
Also [readme.md](https://github.com/hank121314/Defaults/blob/develop/readme.md) and [migration.md](https://github.com/hank121314/Defaults/blob/develop/migration.md) is updated!
When you have a moment, Could you please take a look and let me know if there is anything needs to be changed​.
Thanks 😃 !

readme.md
```diff
@@ -509,7 +558,49 @@ Any `Defaults` key changes made within the closure will not propagate to `Defaul
...
+}
+```
+
+Type: `protocol`
+
+Represents the type after migration.
+
+It should have a associated type name `CodableForm` which protocol conform to `Codable`.
+
+#### `Defaults.CodableType`
```
sindresorhus, line 588: I don't think we need this and `NativeType` in the readme. Just make sure their doc comments are good and link to them from the readme.

readme.md
```diff
@@ -317,6 +300,57 @@ print(UserDefaults.standard.bool(forKey: Defaults.Keys.isUnicornMode.name))
 //=> true
 ```
 
+### Custom types
+
+Although `Defaults` already support many types internal, there might have some situations that you want to use your own type.
+The guide below will show you how to create your own custom type.
```
sindresorhus, line 306: You are not creating a custom type here, you're really showing how to make a custom type work with `Defaults`.

readme.md
```diff
@@ -339,27 +373,72 @@ Create a key with a default value.
...
+Have two associate types `Value` and `Serializable`.
 
-Create a NSSecureCoding key with an optional value.
+`Value` is the type user want to use it.
+
+`Serializable` is the type stored in `UserDefaults`.
+
+`serialize` will be executed before storing to the `UserDefaults` .
+
+`deserialize` will be executed after retrieving its value from the `UserDefaults`.
```
sindresorhus, line 441: I haven't checked, but make sure that all the docs for the types here are also in the source code as doc comments for the types, including examples.

readme.md
```diff
@@ -339,27 +373,72 @@ Create a key with a default value.
...
+public protocol DefaultsSerializable {
+	static var bridge: Defaults.Bridge { get }
+}
 ```
 
-Type: `class`
+Type: `protocol`
 
-Create a NSSecureCoding key with a default value.
+A protocol which can do serialization and de-serialization easily.
```
sindresorhus, line 386: Mention that types needs to conform to this to be supported by `Defaults`.

readme.md
```diff
@@ -339,27 +373,72 @@ Create a key with a default value.
...
 
-The default value is written to the actual `UserDefaults` and can be used elsewhere. For example, with a Interface Builder binding.
+It should have a static variable `bridge` and its protocol should conform to `Defaults.Bridge`.
 
-#### `Defaults.NSSecureCodingOptionalKey` *(alias `Defaults.Keys.NSSecureCodingOptionalKey`)*
+#### `Defaults.CollectionSerializable`
```
sindresorhus, line 390: This should go into an "Advanced" section too.

sindresorhus:
Nice work on the docs. It's looking really good. 👍

sindresorhus:
The docs need to mention the types `Defaults` supports. For example, that it supports Collection, Set, etc, built-in. A list of types would be useful so users have a quick reference.

readme.md
```diff
@@ -509,7 +558,49 @@ Any `Defaults` key changes made within the closure will not propagate to `Defaul
...
+
+```swift
+protocol DefaultsNativeType: Defaults.Serializable {
+	associatedtype CodableForm: Defaults.CodableType
+}
+```
+
+Type: `protocol`
+
+Represents the type after migration.
```
sindresorhus, line 584: It doesn't just represent that. It also means the type is natively supported by userdefaults. I think the description should also describe its purpose.

readme.md
```diff
@@ -509,7 +558,49 @@ Any `Defaults` key changes made within the closure will not propagate to `Defaul
...
+}
+```
+
+Type: `protocol`
+
+Represents the type before migration.
+
+It should have an associated type name `NativeForm` which is the type we want it to store in `UserDefaults`.
+
+And it also have a `toNative()` function to convert itself into `NativeForm`.
```
sindresorhus, line 603: I don't think this type should be in the readme. It's only meant for migration. Users shouldn't have to care about it. It will eventually be removed. I think the source code doc comment should also mention this fact.

Sources/Defaults/Migration/Migration+Defaults.swift
```diff
@@ -1,42 +1,51 @@
...
+			migration(keys, to: version)
+		}
 	}
-	let text = "[\"a\", \"b\", \"c\"]"
-	UserDefaults.standard.set(text, forKey: "array")
 
-	UserDefaults.standard.string(forKey: keyName)
-	//=> ["a","b","c"]
+	/**
+	 Migration the given key's value from json string to `Value`.
```
sindresorhus, line 23: You seem to have an extra one space indention on these doc comments.

migration.md
```diff
@@ -0,0 +1,301 @@
...
+
+extension Defaults.Keys {
+	static let timezone = Defaults.Key<TimeZone?>("TimeZone")
+}
+```
+
+#### Migration steps
+1. Let `TimeZone` protocol conform to `Defaults.Serializable`.
+```swift
+private struct TimeZone: Defaults.Serializable & Codable {
```
sindresorhus, line 55:
```suggestion
private struct TimeZone: Defaults.Serializable, Codable {
```

Same applies to all the others. They are equivalent, but I prefer the comma version.

migration.md
```diff
@@ -0,0 +1,301 @@
...
+	var id: String
+	var name: String
+}
+
+extension Defaults.Keys {
+	static let timezone = Defaults.Key<TimeZone?>("TimeZone")
+}
+```
+
+#### Migration steps
```
sindresorhus, line 52: Always use an empty line after a section

migration.md
```diff
@@ -0,0 +1,301 @@
...
+1. Let `TimeZone` protocol conform to `Defaults.Serializable`.
+```swift
```
sindresorhus, line 54: Always use an empty line before and after a code block.

migration.md
```diff
@@ -0,0 +1,301 @@
+# Migration Guide From v4 to v5
+
+## Summary
```
sindresorhus, line 3: Can you write a short summary of the changes? The below is already good, but it feels more like something you would read after upgrading and seeing compiler errors. It would be nice to have something that explains the changes without going into specifics. Like: v5 changes the internal representation of some types and will require migration. We provide methods to automate the migration process. Etc.  Etc. Basically, explain what is new and what has changed.

sindresorhus: We should also advise devs to be careful and do a lot of manual testing before shipping an update to users.

migration.md
```diff
@@ -0,0 +1,301 @@
...
+```
+3. **Call `Defaults.migration(.period)`**
+4. Now `Defaults[.period]` should be readable.
+
+
+---
+
+### From `Codable` struct in Defaults v4 to `Dictionary` in Defaults v5 (Optional) 
+
+This situation happens when you have a struct which is stored as a codable JSON string before, but now you want it to be stored as a dictionary.
```
sindresorhus, line 220:
```suggestion
This situation happens when you have a struct which is stored as a codable JSON string before, but now you want it to be stored as a native UserDefaults dictionary.
```

sindresorhus (changes requested):

readme.md
```diff
@@ -50,6 +50,26 @@ github "sindresorhus/Defaults"
...
+|      `Double`      |      `[Double]`      |       `Set<Double>`       |              `[T: Double]`               |
+|      `Float`       |      `[Float]`       |       `Set<Float>`        |               `[T: Float]`               |
+|      `String`      |      `[String]`      |       `Set<String>`       |              `[T: String]`               |
+|     `CGFloat`      |     `[CGFloat]`      |      `Set<CGFloat>`       |              `[T: CGFloat]`              |
+|       `Bool`       |       `[Bool]`       |        `Set<Bool>`        |               `[T: Bool]`                |
+|       `Date`       |       `[Date]`       |        `Set<Date>`        |               `[T: Date]`                |
+|       `Data`       |       `[Data]`       |        `Set<Data>`        |               `[T: Data]`                |
+|       `URL`        |       `[URL]`        |        `Set<URL>`         |                `[T: URL]`                |
+| `NSColor` (macOS)  | `[NSColor]` (macOS)  |  `Set<NSColor>` (macOS)   |          `[T: NSColor]` (macOS)          |
+|  `UIColor` (iOS)   |  `[UIColor]` (iOS)   |      `Set<UIColor>`       |           `[T: UIColor]` (iOS)           |
```
sindresorhus, line 68: It feels a bit moot to repeat the basic types we support in each column. I think we can simply say in addition to the single values, we support them wrapped in `Array`, `Set`, `Dictionary`.

This also need to mention support for nested types, for example `[[String: Set<[String: Int]>]]`.

hank121314:
[readme.md](https://github.com/hank121314/Defaults/blob/develop/readme.md) and [migration.md](https://github.com/hank121314/Defaults/blob/develop/migration.md) have been updated!
I also create an internal extension of `Defaults.Serializable` to handle the serialization/deserialization easily. 
As the code below:
https://github.com/hank121314/Defaults/blob/8276a996ec05a1dcd14f394f93155b0245269b2f/Sources/Defaults/Utilities.swift#L158-L200

Thanks for your review 😄 !

sindresorhus:
Tests seems to be failing. Otherwise looks good 👍

hank121314:
> Tests seems to be failing. Otherwise looks good 👍

Hmm...
Fail might cause by testObservePreventPropagationMultipleThreads.
IMO, the duration of usleep is not long enough for DispatchQueue to access it.

hank121314:
Oops misclick 😆

sindresorhus:
@hank121314 This looks really good now. Thanks for your tireless work on this!

Would you be interested in joining the project as a maintainer?

hank121314:
> @hank121314 This looks really good now. Thanks for your tireless work on this!
> 
> Would you be interested in joining the project as a maintainer?

Of course! It's my honor.
Is there any routines I need to do as a maintainer?

sindresorhus:
> Of course! It's my honor.

Awesome! Welcome :)

> Is there any routines I need to do as a maintainer?

Nothing is required, but it would be nice to get help with reviewing issues and PRs. And generally find ways to improve the project.

sindresorhus:
What name do you want me to use for you in the readme?

hank121314:
> Nothing is required, but it would be nice to get help with reviewing issues and PRs. And generally find ways to improve the project.

OK! I will do it 😄 .

> What name do you want me to use for you in the readme?

Hmm.. Let me pick`hank121314` for the name.

PR #73, opened 2021-06-03
Add `Defaults.AnySerializable`

hank121314:
## Summary

`Defaults.AnySerializable` is a type erasure for `Defaults.Serializable`.
It can accept the value conforms to `Defaults.Serializable`.

## Features

With `Defaults.AnySerializable` user can use `Defaults.Key` dynamically.
User will not need to specialize the type of `Defaults.Key`, just use `Defaults.Key<Defaults.AnySerializable>`.

## Usage

```swift
let any = Defaults.Key<Defaults.AnySerializable>("anyKey", default: 121_314)
let int = Defaults[any].get(Int.self)
Defaults[any].set(value: "🦄")
print(Defaults[any].get(String.self)) //=> "🦄"
```

### DIFF
--- a/Sources/Defaults/Defaults+AnySerializable.swift
+++ b/Sources/Defaults/Defaults+AnySerializable.swift
@@ -56,7 +56,7 @@ extension Optional: Defaults.AnySerializableProtocol where Wrapped: Defaults.Any
 	public init<Value: Defaults.Serializable>(_ value: Value) {
 		self = .some(Wrapped(value))
 	}
-	public func get<Value>() -> Value? where Value : DefaultsSerializable {
+	public func get<Value>() -> Value? where Value: DefaultsSerializable {
 		self?.get()
 	}
 	public func get<Value: Defaults.Serializable>(_: Value.Type) -> Value? {
--- a/Sources/Defaults/Defaults+AnySerializable.swift
+++ b/Sources/Defaults/Defaults+AnySerializable.swift
@@ -10,16 +10,18 @@ public protocol DefaultsAnySerializableProtocol: Defaults.Serializable {
 extension Defaults {
 	public typealias AnySerializableProtocol = DefaultsAnySerializableProtocol
 	/**
-	Have an internal property `value` which` should always be a UserDefaults native supported type.
+	Type-erased wrappers for `Defaults.Serializable` values.
+	It can be used when the user wants to create an `Any` value that conforms to `Defaults.Serializable`.
+	It will have an internal property `value` which` should always be a UserDefaults native supported type.
 
-	`get` will deserialize internal value to the type user explicit in function parameter.
+	`get` will deserialize internal value to the type user explicit in the function parameter.
 
 	```
 	let any = Defaults.Key<Defaults.AnySerializable>("independentAnyKey", default: 121_314)
 	print(Defaults[any].get(Int.self)) //=> 121_314
 	```
 
-	- Note: the only way to assign non-serializable value is using `ExpressibleByArrayLiteral` or `ExpressibleByDictionaryLiteral` to assign a type which is not UserDefaults native supported type.
+	- Note: the only way to assign a non-serializable value is using `ExpressibleByArrayLiteral` or `ExpressibleByDictionaryLiteral` to assign a type which is not UserDefaults native supported type.
 
 	```
 	private enum mime: String, Defaults.Serializable {
@@ -49,57 +51,9 @@ extension Defaults {
 		public mutating func set<Value: Serializable>(_ newValue: Value) {
 			value = Value.toSerializable(newValue) ?? ()
 		}
-	}
-}
 
-extension Optional: Defaults.AnySerializableProtocol where Wrapped: Defaults.AnySerializableProtocol {
-	public init<Value: Defaults.Serializable>(_ value: Value) {
-		self = .some(Wrapped(value))
-	}
-	public func get<Value>() -> Value? where Value: DefaultsSerializable {
-		self?.get()
-	}
-	public func get<Value: Defaults.Serializable>(_: Value.Type) -> Value? {
-		self?.get(Value.self)
-	}
-}
-
-extension Defaults {
-	/**
-	Here are the subscription for `Defaults.Key<Defaults.AnySerializable>`.
-
-	By default, `Defaults[any]` will return `Defaults.AnySerializable`.
-	With these subscription we can get the internal `value` with `Defaults[any]`.
-
-	```
-	let any = Defaults.Key<Defaults.AnySerializable>("anyKey", default: 121_314)
-	let int: Int = Defaults[any]
-	print(int) //=> 121_314
-	Defaults[any] = "🦄"
-	let string: String = Defaults[any]
-	print(string) //=> 🦄
-	```
-
-	For more ambiguous cases,  we also provide second parameter `type` to specify the type of `value`
-
-	```
-	let any = Defaults.Key<Defaults.AnySerializable>("anyKey", default: 121_314)
-	Defaults[any, type: Float.self] = 1_213.14
-	let float = Defaults[any, type: Float.self]
-	print(float) //=> 1_213.14
-	```
-	*/
-	public static subscript<T: Serializable, Value: AnySerializableProtocol>(key: Key<Value>) -> T? {
-		get { key.suite[key].get() }
-		set {
-			key.suite[key] = Value(newValue)
-		}
-	}
-
-	public static subscript<T: Serializable, Value: AnySerializableProtocol>(key: Key<Value>, type type: T.Type) -> T? {
-		get { key.suite[key].get(type) }
-		set {
-			key.suite[key] = Value(newValue)
+		public mutating func set<Value: Serializable>(_ newValue: Value, type: Value.Type) {
+			value = Value.toSerializable(newValue) ?? ()
 		}
 	}
 }
@@ -241,6 +195,7 @@ extension Defaults.AnySerializable: ExpressibleByDictionaryLiteral {
 }
 
 extension Defaults.AnySerializable: _DefaultsOptionalType {
+	/// Since nil cannot assign to `Any`, we use `Void` instead of `nil`.
 	public var isNil: Bool { value is Void }
 }
 
--- a/Sources/Defaults/Defaults+AnySerializable.swift
+++ b/Sources/Defaults/Defaults+AnySerializable.swift
@@ -1,14 +1,7 @@
 import CoreGraphics
 import Foundation
 
-public protocol DefaultsAnySerializableProtocol: Defaults.Serializable {
-	init<Value: Defaults.Serializable>(_ value: Value)
-	func get<Value: Defaults.Serializable>() -> Value?
-	func get<Value: Defaults.Serializable>(_: Value.Type) -> Value?
-}
-
 extension Defaults {
-	public typealias AnySerializableProtocol = DefaultsAnySerializableProtocol
 	/**
 	Type-erased wrappers for `Defaults.Serializable` values.
 	It can be used when the user wants to create an `Any` value that conforms to `Defaults.Serializable`.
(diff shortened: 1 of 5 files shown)


sindresorhus:
What's the reason we cannot use a subscript?

Sources/Defaults/Defaults+AnySerializable.swift
```diff
@@ -0,0 +1,210 @@
+import CoreGraphics
+import Foundation
+
+extension Defaults {
+  /**
+  Have an internal property `value` which` should always be a UserDefaults native supported type.
```
sindresorhus, line 6: Use tab-indentation.

Sources/Defaults/Defaults+AnySerializable.swift
```diff
@@ -0,0 +1,210 @@
...
+
+  `get` will deserialize internal value to  the type user explicit in function parameter.
```
sindresorhus, line 8:
```suggestion
  `get` will deserialize internal value to the type user explicit in function parameter.
```

Sources/Defaults/Defaults+AnySerializable.swift
```diff
@@ -0,0 +1,210 @@
...
+
+  ```
+  let any = Defaults.Key<Defaults.AnySerializable>("independentAnyKey", default: 121_314)
+  print(Defaults[any].get(Int.self)) //=> 121_314
+  ```
+
+  `set`will serialize `Value` to UserDefaults native supported type and assign it to internal `value`.
```
sindresorhus, line 15:
```suggestion
  `set` will serialize `Value` to UserDefaults native supported type and assign it to internal `value`.
```

Sources/Defaults/Defaults+AnySerializable.swift
```diff
@@ -0,0 +1,210 @@
...
+  let any = Defaults.Key<Defaults.AnySerializable>("independentAnyKey", default: 121_314)
+
+  // Failed: Attempt to insert non-property list object
+  Defaults[any] = [mime.JSON]
+
+  // Success
+  Defaults[any].set(value: [mime.JSON])
+  ```
+  */
+  public struct AnySerializable: Defaults.Serializable {
```
sindresorhus, line 38:
```suggestion
  public struct AnySerializable: Serializable {
```

sindresorhus (changes requested):

hank121314:
> What's the reason we cannot use a subscript?

Is there any suggest syntax to use a subscript?

Currently, I can change it to

```swift
let any = Defaults.Key<Defaults.AnySerializable>("independentAnyKey", default: 121_314)
print(Defaults[any][Int.self]) //=> 121_314
Defaults[any][String.self] = "🦄"
print(Defaults[any][String.self]) //=> 🦄
```

is it an acceptable changes?

Thank you for your suggestions and code review!

sindresorhus:
> Defaults[any][String.self] = "🦄"

Why does it need `String.self` here? The compiler knows it's a string.

For more ambigious cases, would this work? (Untested)

```swift
Defaults[any] = foo as Foo
```

or

```swift
Defaults[any, type: Foo.self] = foo 
```

hank121314:
> > Defaults[any][String.self] = "🦄"
> 
> Why does it need `String.self` here? The compiler knows it's a string.
> 
> For more ambigious cases, would this work? (Untested)
> 
> ```swift
> Defaults[any] = foo as Foo
> ```
> 
> or
> 
> ```swift
> Defaults[any, type: Foo.self] = foo 
> ```

Yes! Both syntax are available, I will start to work on it.

Thank you for your suggestion 😄!

hank121314:
I think we should have both `Defaults[any][String.self]` and `Defaults[any]`.

Because of this special case:
```swift
enum mime: String, Defaults.Serializable {
	case JSON = "application/json"
}
let magic = Key<[String: Defaults.AnySerializable]>("magic", default: [:])
Defaults[.magic]["enum"] = Defaults.AnySerializable(mime.JSON)
let any_serializable = Defaults[.magic]["enum"]!
XCTAssertEqual(any_serializable[mime.self], mime.JSON)
```

`Defaults[any][String.self]` is a subscription for `Defaults.AnySerializable`.
`Defaults[any]` or `Defaults[any, type: String.self]` is a subscription for `Defaults.Key` which `Defaults.Key<Value> where Value == Defaults.AnySerializable`.

I do not want to public the internal property `value` in `Defaults.AnySerializable`.
So we must have a getter to get the internal `value` in `Defaults.AnySerializable`.

sindresorhus:
Yeah, that's fine. Just make sure you document it well.

hank121314:
I still pick `Defaults[any].get(String.self)` instead of `Defaults[any][String.self]`.
I think `Defaults[any][String.self]` might be a little bit weird, it is not a common use case for subscript.

So now we have a convenience subscript for `Defaults.Key<Defaults.AnySerializable>` which can modify the internal `value` with `Defaults[any]`.
And the `get/set` function in `Defaults.AnySerializable`.
Both of them are documented in [readme.md](https://github.com/hank121314/Defaults/blob/9801f43eaf14f9b83b1e1bd7ba1a8d2a7b30a6e7/readme.md#L642).

Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -297,3 +297,18 @@ extension Defaults {
 		}
 	}
 }
+
+extension Defaults {
+  public struct AnyBridge: Defaults.Bridge {
```
sindresorhus, line 302: Tab-indentation

Sources/Defaults/Defaults+AnySerializable.swift
```diff
@@ -0,0 +1,257 @@
...
+}
+
+extension Defaults.AnySerializable: ExpressibleByDictionaryLiteral {
+	public init(dictionaryLiteral elements: (AnyHashable, Any)...) {
+		self.init(value: [AnyHashable: Any](uniqueKeysWithValues: elements))
+	}
+}
+
+extension Defaults.AnySerializable: _DefaultsOptionalType {
+	public var isNil: Bool { value is Void }
```
sindresorhus, line 244: This could use a doc comment. It's not immediately clear why it's `nil` when it's `Void`.

readme.md
```diff
@@ -625,6 +639,97 @@ Defaults[.setUser].first?.name //=> "Hello"
...
+There might be situations where you want to use `[String: Any]` directly.
+But `Defaults` need its value conforms to `Defaults.Serializable`, so here is a class `Defaults.AnySerializable` to overcome this limitation.
+
+`Defaults.AnySerializable` only available for `Value` which conforms to `Defaults.Serializable`.
+
+#### Primitive type
+
+`Defaults.AnySerializable` conforms to `ExpressibleByStringLiteral`, `ExpressibleByIntegerLiteral`, `ExpressibleByFloatLiteral`, `ExpressibleByBooleanLiteral`, `ExpressibleByNilLiteral`, `ExpressibleByArrayLiteral` and `ExpressibleByDictionaryLiteral`.
+
+So it can assign directly with these primitive types.
```
sindresorhus, line 653: I think we should have a clear warning that the type erasure should only be used when there's no other way to handle it. It has much worse performance.

And don't forget to keep the code documentation comments and readme in sync when making changes.

sindresorhus:
> I still pick Defaults[any].get(String.self) instead of Defaults[any][String.self].

I don't think we should support both though. That's just confusing. I don't really think we need a lot of syntax sugar for such an edge-case either. So let's drop the subscript?

sindresorhus:
Have you considered adding [dynamic member lookup](https://www.hackingwithswift.com/articles/55/how-to-use-dynamic-member-lookup-in-swift) support?

readme.md
```diff
@@ -639,7 +744,7 @@ struct Bag<Element: Defaults.Serializable>: Collection {
 	mutating func insert(element: Element, at: Int) {
 		items.insert(element, at: at)
 	}
-
+P
```
sindresorhus, line 747: ^

readme.md
```diff
@@ -625,6 +639,97 @@ Defaults[.setUser].first?.name //=> "Hello"
...
+```
+
+#### Other types
+
+##### Using `get`, `set`
+
+For other types you will have to assign it like this.
+
+```swift
+enum mime: String, Defaults.Serializable {
```
sindresorhus, line 667: Shouldn't the example use a type that it's not already possible to use without type erasure?

Sources/Defaults/Defaults+AnySerializable.swift
```diff
@@ -0,0 +1,257 @@
...
+public protocol DefaultsAnySerializableProtocol: Defaults.Serializable {
+	init<Value: Defaults.Serializable>(_ value: Value)
+	func get<Value: Defaults.Serializable>() -> Value?
+	func get<Value: Defaults.Serializable>(_: Value.Type) -> Value?
+}
+
+extension Defaults {
+	public typealias AnySerializableProtocol = DefaultsAnySerializableProtocol
+	/**
+	Have an internal property `value` which` should always be a UserDefaults native supported type.
```
sindresorhus, line 13: The first sentence of a type should describe what it is, and then what problem it solves.

sindresorhus (changes requested):

sindresorhus:
This looks pretty good.

The docs needs some work though. If you wanna try fixing some issues yourself, I would recommend just copy-pasting the docs into https://app.grammarly.com It will help with a lot of issues. No worries if not though, I'll do a pass over the docs after merging.

readme.md
hank121314: Every type should be able to use even without the type erasure.
`Defaults.AnySerializable` should only be used in the wrapped types, ex. `Array`, `Set`, `Dictionary`.

hank121314:
> > I still pick Defaults[any].get(String.self) instead of Defaults[any][String.self].
> 
> I don't think we should support both though. That's just confusing. I don't really think we need a lot of syntax sugar for such an edge-case either. So let's drop the subscript?

Yeah! It could be misleading, I have dropped the subscript!

hank121314:
> Have you considered adding [dynamic member lookup](https://www.hackingwithswift.com/articles/55/how-to-use-dynamic-member-lookup-in-swift) support?

Is there any suggest syntax or situation to use the dynamic member lookup?
Can not sure how to add dynamic member lookup in this situation.

Thank you for your suggestion!

hank121314:
> This looks pretty good.
> 
> The docs needs some work though. If you wanna try fixing some issues yourself, I would recommend just copy-pasting the docs into [app.grammarly.com](https://app.grammarly.com) It will help with a lot of issues. No worries if not though, I'll do a pass over the docs after merging.

I have made some fixes on the doc. 
Sorry if I've missed something.
Many thanks for your review!

readme.md
```diff
@@ -625,6 +639,82 @@ Defaults[.setUser].first?.name //=> "Hello"
...
+
+#### Primitive type
+
+`Defaults.AnySerializable` conforms to `ExpressibleByStringLiteral`, `ExpressibleByIntegerLiteral`, `ExpressibleByFloatLiteral`, `ExpressibleByBooleanLiteral`, `ExpressibleByNilLiteral`, `ExpressibleByArrayLiteral` and `ExpressibleByDictionaryLiteral`.
+
+So it can assign directly with these primitive types.
+
+```swift
+let any = Defaults.Key<Defaults.AnySerializable>("anyKey", default: 1)
+Defaults[any] = "🦄"
```
sindresorhus, line 659: Isn't this outdated now?

sindresorhus:
Dynamic member lookup would have let users do `Defaults[any].count` instead of `Defaults[any].get().count`, but I no longer think it's worth adding sugar for this either.

hank121314:
> Dynamic member lookup would have let users do `Defaults[any].count` instead of `Defaults[any].get().count`, but I no longer think it's worth adding sugar for this either.

Ok! I think the docs are clear enough to let the user use `get/set` functions easily without adding sugar.
Except for the docs, I think this pr is ready to merge?
Thank you once again for passing over the docs!

sindresorhus:
Nice work!

Do you plan any more PRs in the near-term or should I do a new release with this now?

hank121314:
> Nice work!
> 
> Do you plan any more PRs in the near-term or should I do a new release with this now?

I would like to create one more PR for #78. 
Support SwiftUI Color will be a huge improvement for those developers using `Defaults` in SwiftUI.

sindresorhus:
> I would like to create one more PR for #78.

👍

PR #76, opened 2021-06-16
Add `.defaultValue` and `.isDefaultValue` to `@Default`

sindresorhus:
I need it to disable an input the value matches the default value. This is what I'm currently using:

```swift
.disabled(textSize == Defaults.Keys.textSize.defaultValue)
```

### DIFF
--- a/Sources/Defaults/SwiftUI.swift
+++ b/Sources/Defaults/SwiftUI.swift
@@ -87,6 +87,9 @@ public struct Default<Value: Defaults.Serializable>: DynamicProperty {
 
 	public var projectedValue: Binding<Value> { $observable.value }
 
+	/// The default value of the key.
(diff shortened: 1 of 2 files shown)


sindresorhus:
I wonder if it also makes sense to add `.isDefaultValue` for convenience.

hank121314:
> I wonder if it also makes sense to add `.isDefaultValue` for convenience.

Yeah! It will be great if we have `.isDefaultValue` 😄.
But `Defaults.Serializable` is not `Equatable`, we should document that only `Value` conforms to `Defaults.Serializable & Equatable` can use `.isDefaultValue`.

sindresorhus:
> we should document that only Value conforms to Defaults.Serializable & Equatable can use .isDefaultValue.

I think that's made clear through the type signature.

Issue #79, opened 2021-09-10
Force serialization to not use `Codable`?


sindresorhus:
It would be useful to have a way to force the serialization to not use the `Codable` implementation.

Let's say I have an enum:

```swift
enum Foo: String, CaseIterable, Codable, Defaults.Serializable {
	case bar
}
```

It needs to stay `Codable` as I need to be able to use it for some other purposes too, where `Codable` is needed. This also applies to types you don't control that are already `Codable`.

However, I would like to store it using the native enum handling instead of using `Codable` (so it's not wrapped in double-quotes). It would be nice if we had a way to force that.

This would also be a nice safe-guard. Imagine you have a type that is `Defaults.Serializable`, but not `Codable`, and later on you add `Codable` conformance to it, then the stored representation would change and you could accidentally lose the data.

@@hank121314 Thoughts?

hank121314:
This might be the same feature as you mentioned in #80?
> It would be nice if a user could somehow choose whether to use the Codable or NSSecureCoding for the serialization.

sindresorhus:
> This might be the same feature as you mentioned in #80?

Yes, just wanted to point out another scenario where it would have been useful.

hank121314:
Maybe we can try to add a `usingCodable` property in `Defaults.Key`?

For example.
```
let key = Defaults.Key<Foo>(default: Foo.bar, usingCodable: false)
```

I am working on this will create a PR later!

sindresorhus:
I was thinking it would be in the type system:

```swift
enum Foo: String, CaseIterable, Codable, Defaults.Serializable {
	case bar
}

// And in my code:
extension Foo: Defaults.NoCodable {}

extension Defaults.Keys {
	static let foo = Key<Foo>("foo", default: .bar)
}
```

The benefit of your idea is that it applies per key and not per type, but mine is definitely simpler, and I cannot think of many situations where you want it per key and not just per type.

And if you do need it per key, could something like this work?

```swift
enum Foo: String, CaseIterable, Codable, Defaults.Serializable {
	case bar
}

// And in my code:
typealias FooNoCodable = Foo & Defaults.NoCodable

extension Defaults.Keys {
	static let foo = Key<FooNoCodable>("foo", default: .bar)
}
```

If `Defaults.NoCodable` has no requirements, it could be used as a "marker protocol", but I'm honestly not sure if `Foo` would still be assignable to it.

I'm not saying mine is better, but it would be good to discuss this a bit, to explore our options.

[...]

PR #82, opened 2021-09-16
Add support to not use `Codable`

hank121314:
## Summary

This PR fixed: #79.
I have added a `usingCodable` property to `Defaults.Key`.
If users set `usingCodable` to false, so the serialization will not use `Codable`.

## Usage

```swift
private enum Foo: Int, CaseIterable, Codable, Defaults.Serializable {
	case zero
	case one
}

let key = Defaults.Key<Foo>(keyName, default: fixture, usingCodable: false)
```

### DIFF
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -20,7 +20,6 @@ public enum Defaults {
 	public typealias CollectionSerializable = DefaultsCollectionSerializable
 	public typealias SetAlgebraSerializable = DefaultsSetAlgebraSerializable
 	public typealias Bridge = DefaultsBridge
-	typealias CodableBridge = DefaultsCodableBridge
 
 	public class Keys: BaseKey {
 		public typealias Key = Defaults.Key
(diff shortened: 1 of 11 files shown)


Sources/Defaults/Defaults.swift
```diff
@@ -36,10 +35,12 @@ public enum Defaults {
 
 	public final class Key<Value: Serializable>: AnyKey {
 		public let defaultValue: Value
+		/// A flag to determine whether CodableBridge is a first-class serializator
+		var usingCodable = true
 
 		/// Create a defaults key.
 		/// The `default` parameter can be left out if the `Value` type is an optional.
-		public init(_ key: String, default defaultValue: Value, suite: UserDefaults = .standard) {
+		public init(_ key: String, default defaultValue: Value, suite: UserDefaults = .standard, usingCodable: Bool = true) {
```
sindresorhus, line 43: I don't the parameter should be available when the value is not `Codable`. It wouldn't work anyway.

Sources/Defaults/Defaults.swift
```diff
@@ -63,6 +83,13 @@ public enum Defaults {
 			key.suite[key] = newValue
 		}
 	}
+
+	public static subscript<Value: Serializable & Codable>(key: Key<Value>) -> Value {
+		get { key.suite[key] }
+		set {
+			key.suite[key] = newValue
+		}
+	}
```
sindresorhus, line 92: What's the reason for this overload? The logic is the same, so I don't really see what it adds.

Sources/Defaults/Observation.swift
```diff
@@ -304,6 +323,19 @@ extension Defaults {
 		return observation
 	}
 
+	public static func observe<Value: Serializable & Codable>(
```
sindresorhus, line 326: I don't think `observe` should need to know about the extra `Codable` logic. It feels weird having to add overloads for methods like this. The logic should be internal in the serialization.

hank121314:
Close this since we find a better solution https://github.com/sindresorhus/Defaults/issues/79#issuecomment-926688704.

Issue #98, opened 2022-05-06
Support `ClosedRange` and `Range`


sindresorhus:
I think they should just serialize into an array with two elements: `.lowerBound` and `.upperBound`.

@hank121314 Thoughts?

hank121314:
Yes! swift also only encode these two elements. 
I think we might also support any `Bound` that conforms to `Defaults.Serializable`?

Reference: https://github.com/apple/swift/blob/223d73cc470d2395b3d55d07eaf6866e7f6171f9/stdlib/public/core/ClosedRange.swift#L495-L501

sindresorhus:
> I think we might also support any Bound that conforms to Defaults.Serializable?

👍

Issue #101, opened 2022-05-06
Defaults crashes on custom bridge with array of serializable values


sindresorhus:
```swift
struct PlainHourMinuteTimeRange: Hashable, Codable {
	var start: PlainHourMinuteTime
	var end: PlainHourMinuteTime
}

extension PlainHourMinuteTimeRange: Defaults.Serializable {
	struct Bridge: Defaults.Bridge {
		typealias Value = PlainHourMinuteTimeRange
		typealias Serializable = [PlainHourMinuteTime]

		public func serialize(_ value: Value?) -> Serializable? {
			guard let value = value else {
				return nil
			}

			return [value.start, value.end]
		}

		public func deserialize(_ object: Serializable?) -> Value? {
			guard
				let array = object,
				let start = array[safe: 0],
				let end = array[safe: 1]
			else {
				return nil
			}

			return .init(start: start, end: end)
		}
	}

	static let bridge = Bridge()
}

struct PlainHourMinuteTime: Hashable, Codable, Defaults.Serializable {
	var hour: Int
	var minute: Int
}

extension Collection {
	subscript(safe index: Index) -> Element? {
		indices.contains(index) ? self[index] : nil
	}
}
```

Error:

```
022-05-06 15:55:56.549366+0700 Dato[75451:1229122] [User Defaults] Attempt to set a non-property-list object (
    "Dato.PlainHourMinuteTime(hour: 9, minute: 0)",
    "Dato.PlainHourMinuteTime(hour: 22, minute: 0)"
) as an NSUserDefaults/CFPreferences value for key hourlyChimeTimeRange
```

@hank121314 I think we may need to check whether the value is serializable in the array bridge and if so, use its serializer. I'm not sure though. I might be doing something wrong too. However, we should never let it hard crash.

hank121314:
By using the custom bridge, `Defaults` will only use this bridge to serialize. Type `Serializable` in custom bridge should be `UserDefaults` native supported type.  But I think this might be a bit awkward. 

> I think we may need to check whether the value is serializable in the array bridge and if so, use its serializer.

This is a great way to improve this kind of situation. I will try to implement it 😃 !

PR #102, opened 2022-05-11
Add support for `ClosedRange` and `Range`

hank121314:
## Summary

Fixes: #98.
Create a `RangeBridge` to support `ClosedRange` and `Range` serialization.
`RangeBridge` can serialize the object which protocol conforms to `Defaults.RangeSerializable`.
`Defaults.RangeSerializable` is a protocol that contains some essential properties for serializing and deserializing `ClosedRange` and `Range`.

### DIFF
--- a/Sources/Defaults/Defaults+Bridge.swift
+++ b/Sources/Defaults/Defaults+Bridge.swift
@@ -347,6 +347,7 @@ extension Defaults {
 			guard let value = value else {
 				return nil
 			}
+
 			if Bound.isNativelySupportedType {
 				return [value.lowerBound, value.upperBound]
 			}
@@ -365,6 +366,7 @@ extension Defaults {
 			guard let object = object else {
 				return nil
 			}
+
 			if Bound.isNativelySupportedType {
 				guard
 					let lowerBound = object[safe: 0] as? Bound,
--- a/Sources/Defaults/Defaults+Bridge.swift
+++ b/Sources/Defaults/Defaults+Bridge.swift
@@ -338,6 +338,57 @@ extension Defaults {
 }
 
 extension Defaults {
+	public struct RangeBridge<T: RangeSerializable>: Bridge {
+		public typealias Value = T
+		public typealias Serializable = [Any]
+		typealias Bound = T.Bound
+
+		public func serialize(_ value: Value?) -> Serializable? {
+			guard let value = value else {
+				return nil
+			}
+			if Bound.isNativelySupportedType {
+				return [value.lowerBound, value.upperBound]
+			}
+
+			guard
+				let lowerBound = Bound.bridge.serialize(value.lowerBound as? Bound.Value),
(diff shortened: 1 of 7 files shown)


Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -337,6 +337,57 @@ extension Defaults {
...
+	public struct RangeBridge<T: RangeSerializable>: Bridge {
+		public typealias Value = T
+		public typealias Serializable = [Any]
+		typealias Bound = T.Bound
+
+		public func serialize(_ value: Value?) -> Serializable? {
+			guard let value = value else {
+				return nil
+			}
+			if Bound.isNativelySupportedType {
```
sindresorhus, line 350: Shouldn't this be handled in the `Bound.bridge.serialize` method implementation?

readme.md
```diff
@@ -96,6 +96,8 @@ Add `https://github.com/sindresorhus/Defaults` in the [“Swift Package Manager
 - `Color` (SwiftUI)
 - `Codable`
 - `NSSecureCoding`
+- `Range`
+- `ClosedRange`
 
 Defaults also support the above types wrapped in `Array`, `Set`, `Dictionary`, and even wrapped in nested types. For example, `[[String: Set<[String: Int]>]]`.
```
sindresorhus, line 102: Range and ClosedRange should be mention in this sentence instead.

sindresorhus (changes requested):

Sources/Defaults/Defaults+Bridge.swift
hank121314: `Bound.bridge` depends on `Bound` type. 
If `Bound.bridge` is a user custom bridge, we should not let them handle this. 
This situation seems to be similar to #101.

https://github.com/sindresorhus/Defaults/blob/3535f3d088113cf24705014eec6a17f0fd73237f/Sources/Defaults/Utilities.swift#L201-L211

`toSerializable` will only check once. I think we should make it recursively.

Tests/DefaultsTests/DefaultsRangeTests.swift
```diff
@@ -0,0 +1,221 @@
+import Foundation
+import Defaults
+import XCTest
+
+private struct Date {
```
sindresorhus, line 5: I would name it `CustomDate` so it's not confused with the built-in `Date` when reading the code.

Sources/Defaults/Defaults+Bridge.swift
sindresorhus: > toSerializable will only check once. I think we should make it recursively.

👍

sindresorhus (changes requested):

sindresorhus:
https://github.com/sindresorhus/Defaults/pull/102#discussion_r870015553

hank121314:
> 🙌

Thanks for the quick review! As for #101, I will make a separate PR to solve this issue!

Issue #97, opened 2022-04-20
Color space is not preserved for `Color` type on iOS


sindresorhus:
It is however preserved for `UIColor`.

Report: https://twitter.com/jordibruin/status/1516728958695354369

I have added some tests and this one fails on iOS: https://github.com/sindresorhus/Defaults/commit/3535f3d088113cf24705014eec6a17f0fd73237f#diff-d780be73215173c5c7564ee04feb252d30d24c3842b4d9f4aada400be8f37352R21

This seems to be a bug in when SwiftUI translates between `UIColor` and `Color`?

sindresorhus:
@hank121314 Any ideas?

hank121314:
The behavior of `UIColor` seems weird.

I had tested with the following code

```swift
let uiColor = UIColor(displayP3Red: 1, green: 0.3, blue: 0.7, alpha: 1)
print(uiColor.cgColor.colorSpace!) // sRGB
let p3 = CGColorSpace(name: CGColorSpace.displayP3)!
let cgColor = uiColor.cgColor.converted(to: p3, intent: .defaultIntent, options: nil)!
print(cgColor.colorSpace!) // P3
let fixture = UIColor(cgColor: cgColor)
print(fixture.cgColor.colorSpace!) // P3
```

Even if we create `UIColor` with `init(displayP3Red:green:blue:alpha:)` the color space is still `sRGB`.
And also when using `NSSecureCodingBridge` to serialize and deserialize `UIColor`, color space will also return to `sRGB`.
There is need for further investigation 🤓 .

sindresorhus:
`UIColor` always converts to and stores the color internally as "extended sRGB", but I had thought that it would store the original color space.

Maybe we could subclass `UIColor` and add a color space property. I'm not sure whether that would be archived or not though.

hank121314:
It is really hard to add a color space property in `UIColor` subclass. Reference: [You Can't Subclass UIColor](https://cjwirth.com/tech/you-cant-subclass-uicolor).
Is it acceptable to create a class like this?

```swift
public class DefaultsUIColor {
	private var _colorSpace: Color.RGBColorSpace = .sRGB
	public var colorSpace: Color.RGBColorSpace { return _colorSpace }
	
	public let color: UIColor
	public init(_ colorSpace: Color.RGBColorSpace, red: Double, green: Double, blue: Double, alpha: Double)
}
```

sindresorhus:
[...]

PR #105, opened 2022-06-06
Fix preserve color space for `Color` type

hank121314:
## Summary

Fixes: #97.

By default, `ColorBridge` will convert `Color` into `UIColor` in non-macOS platform.
`UIColor` does not preserve the color space, but `Color` should store the color space.

## Solution

Trying to preserve color space whenever `Color.cgColor` exists.
If `SwiftUi.Color` has `cgColor` property, Defaults will serialize and store its colorspace and components, then deserialize it with these properties.
If not, Defaults will use `NSSecureCodingBridge` to do the serialization and deserialization.

## Thanks

### DIFF
--- a/Sources/Defaults/Defaults+Bridge.swift
+++ b/Sources/Defaults/Defaults+Bridge.swift
@@ -442,11 +442,11 @@ extension Defaults {
 				return nil
 			}
 
-			#if os(macOS)
-			return Value(cgColor)
-			#else
-			return Value(cgColor: cgColor)
-			#endif
+			if #available(macOS 12.0, macOSApplicationExtension 12.0, *) {
+				return Value(cgColor: cgColor)
+			} else {
+				return Value(cgColor)
+			}
 		}
 	}
 }
--- a/Sources/Defaults/Defaults+Bridge.swift
+++ b/Sources/Defaults/Defaults+Bridge.swift
@@ -394,10 +394,7 @@ extension Defaults {
 	/**
 	The bridge which is responsible for `SwiftUI.Color` serialization and deserialization.
 
-	It is unsafe to convert `SwiftUI.Color` to `UIColor` and use `UIColor.bridge` to serialize it.
-	Because `UIColor` does not hold a color space, but `Swift.Color` does(which means color space might get lost in the conversion).
-	The bridge will always trying to preserve color space whenever `Color.cgColor` exists.
-	Only if `Color.cgColor` is `nil`, it will use `UIColor.bridge` to do serialization and deserialization.
+	It is unsafe to convert `SwiftUI.Color` to `UIColor` and use `UIColor.bridge` to serialize it, because `UIColor` does not hold a color space, but `Swift.Color` does (which means color space might get lost in the conversion). The bridge will always try to preserve the color space whenever `Color#cgColor` exists. Only when `Color#cgColor` is `nil`, will it use `UIColor.bridge` to do the serialization and deserialization.
 	*/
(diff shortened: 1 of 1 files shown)


Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -406,14 +406,42 @@ extension Defaults {
...
-			guard let nativeColor = NativeColor.bridge.deserialize(object) else {
+			if let object = object as? NativeColor.Serializable {
+				guard let nativeColor = NativeColor.bridge.deserialize(object) else {
+					return nil
+				}
+
+				return Value(nativeColor)
+			}
+
+			guard
```
sindresorhus, line 429: Can you add a short comment that says why we do this. Because `UIColor` does not hold a color space, but `Swift.Color` does. Just for any readers wondering why we made it so complicated.

hank121314: Updated! Thanks for the reminder!

Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -406,14 +414,42 @@ extension Defaults {
...
+			}
+			#if os(macOS)
+			guard let nativeColor = NativeColor(cgColor: cgColor) else {
 				return nil
 			}
+			#else
+			let nativeColor = NativeColor(cgColor: cgColor)
+			#endif
 
 			return Value(nativeColor)
```
sindresorhus, line 453:
```suggestion
			return Value(cgColor)
```

`Swift.Color` accepts CGColor directly.

sindresorhus (changes requested):

Sources/Defaults/Defaults+Bridge.swift
hank121314: `ColorBridge` is available for macOS 11.0.
But `SwiftUI.Color:init(cgColor:)` is only available in macOS 12.0 or newer 😢 .

sindresorhus:
After this PR is merged, do you plan any more changes, or should I do a release?

hank121314:
> After this PR is merged, do you plan any more changes, or should I do a release?

I have no immediate plan for any more changes.
I think we should do a release after this PR gets merged.

Thanks!

Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -407,15 +412,45 @@ extension Defaults {
...
+				let components = object[1] as? [CGFloat],
+				let cgColor = CGColor(colorSpace: colorspace, components: components)
+			else {
+				return nil
+			}
+
+			#if os(macOS)
+			guard let nativeColor = NativeColor(cgColor: cgColor) else {
 				return nil
 			}
```
sindresorhus, line 448: This can be removed too. Just pass cgColor directly on macOS too.

hank121314: Many thanks for the reminder!
Getting confused between the availability of `init(_ cgColor:)` and `init(cgColor:)`:sweat_smile:.

Sources/Defaults/Defaults+Bridge.swift
```diff
@@ -407,15 +412,41 @@ extension Defaults {
...
+			else {
 				return nil
 			}
 
-			return Value(nativeColor)
+			#if os(macOS)
+			return Value(cgColor)
+			#else
+			return Value(cgColor: cgColor)
+			#endif
```
sindresorhus, line 449: Nitpick, but it would be better to use an available check:

```suggestion
			if #available(iOS 15, macOS 12, tvOS 15, watchOS 8, *) {
				return Value(cgColor: cgColor)
			} else {
				return Value(cgColor)
			}
```

PR #114, opened 2022-09-27
Replace standard library primary associated types with conformance

hank121314:
## Summary

Xcode 14 has the macOS 12.3 SDK, where this standard library does not support primary associated types.

Reference: https://forums.swift.org/t/xcode-14-rc-cannot-specialize-protocol-type/60171/22

## Thanks

### DIFF
--- a/Sources/Defaults/Migration/v5/Migration+Extensions.swift
+++ b/Sources/Defaults/Migration/v5/Migration+Extensions.swift
@@ -198,7 +198,7 @@ extension Defaults.SetAlgebraSerializable where Self: Defaults.NativeType, Eleme
 	public typealias CodableForm = [Element.CodableForm]
 }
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Yeah, I assumed I wouldn't do another release until Xcode 14.1 is out, which would contain the macOS 13 SDK. It should be within the next 2-3 weeks.

I can merge this if we need to do a release before then though.

hank121314:
> Yeah, I assumed I wouldn't do another release until Xcode 14.1 is out, which would contain the macOS 13 SDK. It should be within the next 2-3 weeks.

Got it! Many thanks for the explanation!

> I can merge this if we need to do a release before then though.

I am trying to fix #101. 
Since that bug is not that critical, is it worth doing a release?

sindresorhus:
Closing as Xcode 14.1 RC is out and final will be out in a few days.

Issue #108, opened 2022-08-07
Support dynamic default value?


sindresorhus:
Sometimes you cannot define a static default value as it may change during the lifetime of the app. For example, the default value is the current system appearance or a connected device (which may disconnect).

For situations like these, it would be nice if the default value could also be a getter:

```swift
extension Defaults.Keys {
	static let camera = Key<AVCaptureDevice?>("camera") { .default(for: .video) }
}
```

The getter would be executed every time `Defaults[.camera]` is called.

sindresorhus:
@hank121314 Thoughts?

hank121314:
I think we might be able to create a closure type `Getter`, ex. `typealias Getter = (Value) -> Value`.
The function parameter of `Getter` will be a deserialized `Value` from `UserDefaults`.
And add property `getter` with `Getter` type in `Defaults.Key`.

sindresorhus:
Sounds good to me 👍

hank121314:
I think we might enable this feature only if `Value` is optional(`Value?`)?

If `Value` is not optional, user will need to set `defaultValue` which will be used in `UserDefaults.register`.
If `defaultValue` have been registered, `UserDefaults._get` will always get a `Value` which will not trigger `defaultValue`.

sindresorhus:
> If Value is not optional, user will need to set defaultValue which will be used in UserDefaults.register.

`UserDefaults.register` was useful for legacy use-cases outside of Defaults, but it's not that useful these days. There's not really any good reason to use `UserDefaults` directly. So I think we can just document that when you use the getter, the default is not set in `UserDefaults.register`.

PR #118, opened 2022-10-19
Support serialize and deserialize nested custom types

hank121314:
## Summary

This PR fixes: #101.

`Defaults` now support serializing and deserializing nested custom types.

## Thanks

### DIFF
--- a/Tests/DefaultsTests/DefaultsCustomBridgeTests.swift
+++ b/Tests/DefaultsTests/DefaultsCustomBridgeTests.swift
@@ -38,12 +38,56 @@ public final class DefaultsUserBridge: Defaults.Bridge {
 
 private let fixtureCustomBridge = User(username: "hank121314", password: "123456")
 
+struct PlainHourMinuteTimeRange: Hashable, Codable {
+	var start: PlainHourMinuteTime
+	var end: PlainHourMinuteTime
+}
+
+extension PlainHourMinuteTimeRange: Defaults.Serializable {
+	struct Bridge: Defaults.Bridge {
+		typealias Value = PlainHourMinuteTimeRange
+		typealias Serializable = [PlainHourMinuteTime]
+
+		public func serialize(_ value: Value?) -> Serializable? {
+			guard let value = value else {
+				return nil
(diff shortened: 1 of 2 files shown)


Tests/DefaultsTests/DefaultsCustomBridgeTests.swift
```diff
@@ -148,6 +192,35 @@ final class DefaultsCustomBridge: XCTestCase {
...
+		let key = Defaults.Key<PlainHourMinuteTimeRange>("independentCustomBridgeRecursiveKey", default: range)
+		XCTAssertEqual(Defaults[key].start.hour, range.start.hour)
+		XCTAssertEqual(Defaults[key].start.minute, range.start.minute)
+		XCTAssertEqual(Defaults[key].end.hour, range.end.hour)
+		XCTAssertEqual(Defaults[key].end.minute, range.end.minute)
+		guard let rawValue = UserDefaults.standard.array(forKey: key.name) as? [String] else {
+			XCTFail("rawValue should not be nil")
+			return
+		}
+		XCTAssertEqual(rawValue, ["{\"minute\":0,\"hour\":1}", "{\"minute\":0,\"hour\":2}"])
```
sindresorhus, line 208: You can use `#"x"#` string to avoid the escaping.

sindresorhus (changes requested):

sindresorhus:
Thanks for fixing this. Do you plan more changes after this PR? If not, I will make a release.

hank121314:
> Do you plan more changes after this PR?

I would like to move on to #108.
But I think this feature might be useful, maybe it is worth to make a minor release?

sindresorhus:
> maybe it is worth to make a minor release?

Ah, then it's no rush. We can wait until after that. Another reason to wait a little bit more is that the release would require Xcode 14.1.

hank121314:
> Ah, then it's no rush. We can wait until after that. Another reason to wait a little bit more is that the release would require Xcode 14.1.

Got it! That makes sense. Thanks for the explanation!

PR #121, opened 2022-11-15
Support dynamic default value

hank121314:
## Summary

Fixes: #108 .

`Defaults` now accept default value as a `getter`.

## Approach

Replace default value's type with type `typealias DefaultGetter = () -> Value`.
And change `Key.defaultValue` to a computed property which getter is `DefaultGetter`.

## Thanks

Sorry for the delay of this PR and thanks for your code review 😄 !

### DIFF
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -42,12 +42,13 @@ public enum Defaults {
 
 	public final class Key<Value: Serializable>: AnyKey {
 		/**
-		The `defaultValueGetter` will be executed in these situations:
+		It will be executed in these situations:
 
 		- `UserDefaults.object(forKey: string)` return `nil`
 		- `bridge` cannot deserialize `Value` from `UserDefaults`
-		 */
+		*/
 		private let defaultValueGetter: () -> Value
+
 		public var defaultValue: Value { defaultValueGetter() }
 
 		/**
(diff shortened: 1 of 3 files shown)


Sources/Defaults/Defaults.swift
```diff
@@ -41,15 +41,24 @@ public enum Defaults {
...
+		public var defaultValue: Value { defaultGetter() }
 
 		/**
 		Create a defaults key.
 
 		The `default` parameter can be left out if the `Value` type is an optional.
 		*/
-		public init(_ key: String, default defaultValue: Value, suite: UserDefaults = .standard) {
-			self.defaultValue = defaultValue
+		public init(_ key: String, default defaultGetter: @autoclosure @escaping DefaultGetter, suite: UserDefaults = .standard) {
```
sindresorhus, line 59: I don't think it should be an autoclosure. It will mean that people can accidentally pass a function. I think this parameter should remain unchanged and just wrapped it in a closure internally.

Sources/Defaults/Defaults.swift
```diff
@@ -41,15 +41,24 @@ public enum Defaults {
...
 	public final class Key<Value: Serializable>: AnyKey {
-		public let defaultValue: Value
+		public typealias DefaultGetter = () -> Value
+		/**
+		The `defaultGetter` will be executed in these situations:
+
+		- `UserDefaults.object(forKey: string)` return `nil`
+		- `bridge` cannot deserialize `Value` from `UserDefaults`
+		 */
+		private let defaultGetter: DefaultGetter
```
sindresorhus, line 51:
```suggestion
		private let defaultValueGetter: DefaultGetter
```

sindresorhus:
> Sorry for the delay of this PR and thanks for your code review 😄 !

No need to be sorry. This is open source. We do this stuff for fun when we can.

hank121314:
> No need to be sorry. This is open source. We do this stuff for fun when we can.

Got it! Well said! Working on this stuff really makes me fun.

I have applied some changes, thanks for your code review once again!

hank121314:
> By the way, we now have API docs: [swiftpackageindex.com/sindresorhus/defaults/main/documentation/defaults/defaults](https://swiftpackageindex.com/sindresorhus/defaults/main/documentation/defaults/defaults)

This is awesome! It is much better than showing API in `readme.md`! Thanks!

sindresorhus:
✨

PR #123, opened 2022-11-17
Remove `DefaultsBaseKey`

sindresorhus:
Just realized we don't really need it.

@hank121314 Can you think of any reason to keep it?


hank121314:
I think `AnyKey` is enough to use. 
Can not figure out the use-case that need to declare `name: string` and `suite: UserDefaults` as protocol.

Issue #120, opened 2022-11-08
Validate the default key name


sindresorhus:
In the next major version, we should strictly validate the key name to ensure users don't accidentally make the key's unable to be observed.

The rules are as follows:

> The key must be ASCII, not start with `@`, and cannot contain a dot (`.`).

hank121314:
I think we might have three ways to do it.

1. Mark `Defaults.Key` initializer as `throws`:
- pros:
  1. Can throw an error early when user input an invalid name.
- cons:
  1. Need to add `try` when initialize `Defaults.Key`.

2. Add an assertion in `Defaults.Key` initializer:
- pros:
  1. No need to add `try`.
- cons:
  1. Assertion is not available in release mode.

3. Mark `KVO` related function(ex. `Defaults.observe`) as `throws`:
- pros:
  1. No need to add `try`.

sindresorhus:
Only 2. is really feasible.

The question is, do assertions work when they are in a package and you build your app in debug mode.

hank121314:
> Actually, a better solution could be to show a warning in Xcode using this hack: [pointfreeco/swift-composable-architecture@main/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift](https://github.com/pointfreeco/swift-composable-architecture/blob/main/Sources/ComposableArchitecture/Internal/RuntimeWarnings.swift?rgh-link-date=2022-11-24T11%3A12%3A49Z)

This looks awesome! Will try to implement this in `Defaults`.

PR #126, opened 2022-12-02
Validate the default key name and emit with Xcode runtime warning

hank121314:
## Summary

This PR fixes: #120 

1. Implement Xcode runtime warning in `Defaults`.
   In order to let `Defaults.Key.init` support `@_alwaysEmitIntoClient`, 
   there is a needed to mark some functions/properties as `@usableFromInline`. 
   And I also explicit the `internal` keyword for these functions/properties(To remind the developer not to change the access level).

2. Validate the default key name and if it is wrong, emit with Xcode runtime warning.

## Screenshot

screenshot-xcode-runtime-warning

## Thanks

### DIFF
--- a/Sources/Defaults/Utilities.swift
+++ b/Sources/Defaults/Utilities.swift
@@ -167,7 +167,7 @@ extension Collection {
 extension Defaults {
 	@usableFromInline
 	internal static func isValidKeyPath(name: String) -> Bool {
-		// Not start with `@`, Must be ASCII and cannot contain a dot (`.`).
+		// The key must be ASCII, not start with @, and cannot contain a dot.
 		return !name.starts(with: "@") && name.allSatisfy { $0 != "." && $0.isASCII }
 	}
 }
--- a/Sources/Defaults/Utilities.swift
+++ b/Sources/Defaults/Utilities.swift
@@ -167,16 +167,8 @@ extension Collection {
 extension Defaults {
 	@usableFromInline
 	internal static func isValidKeyPath(name: String) -> Bool {
-		// Not start with `@`
-		return !name.starts(with: "@") && name.allSatisfy {
-			// Must be ASCII
-			guard let asciiValue = $0.asciiValue else {
-				return false
-			}
-
-			// Cannot contain a dot (`.`)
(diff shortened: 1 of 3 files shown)


Sources/Defaults/Utilities.swift
```diff
@@ -160,6 +164,21 @@ extension Collection {
...
+	internal static func isValidKeyPath(name: String) -> Bool {
+		// Not start with `@`
+		return !name.starts(with: "@") && name.allSatisfy {
+			// Must be ASCII
+			guard let asciiValue = $0.asciiValue else {
+				return false
+			}
+
+			// Cannot contain a dot (`.`)
+			return asciiValue != 46
```
sindresorhus, line 178: I think `{ $0 != "." && $0.isASCII }` would be simpler.

Sources/Defaults/Utilities.swift
```diff
@@ -217,3 +237,52 @@ extension Defaults.Serializable {
...
 }
+
+#if DEBUG
+/**
+Get SwiftUI dynamic shared object.
+
+Reference: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html
+*/
+@usableFromInline
+internal let dynamicSharedObject = { () -> UnsafeMutableRawPointer in
```
sindresorhus, line 248:
```suggestion
internal let dynamicSharedObject: UnsafeMutableRawPointer = {
```

sindresorhus (changes requested):

sindresorhus:
> In order to let Defaults.Key.init support @_alwaysEmitIntoClient,

Why does `Defaults.Key.init` need to support `@_alwaysEmitIntoClient`?

hank121314:
> Why does Defaults.Key.init need to support @_alwaysEmitIntoClient?

With `@_alwaysEmitIntoClient`, it will let runtime warning appear at call site.

Here is the screenshot if not using `@_alwaysEmitIntoClient`:

screenshot-without-alwaysEmitIntoClient

sindresorhus:
Nice work 👍

PR #147, opened 2023-09-02
Fixes infinite recursion when casting `AnySerializable` to wrong type

leoMehlig:
If a values associated type `Serializable` , equals the values type, `toValue` would infinitely call itself.

The test `testWrongCast` reproduces this and I fixed it by comparing the `nextType` with the current type.

### DIFF
--- a/Sources/Defaults/Utilities.swift
+++ b/Sources/Defaults/Utilities.swift
@@ -194,7 +194,7 @@ extension Defaults.Serializable {
 			return anyObject
 		}
 
-		guard let nextType = T.Serializable.self as? any Defaults.Serializable.Type else {
+        guard let nextType = T.Serializable.self as? any Defaults.Serializable.Type, nextType != T.self else {
 			// This is a special case for the types which do not conform to `Defaults.Serializable` (for example, `Any`).
 			return T.bridge.deserialize(anyObject as? T.Serializable) as? T
 		}
--- a/Sources/Defaults/Utilities.swift
+++ b/Sources/Defaults/Utilities.swift
@@ -194,7 +194,7 @@ extension Defaults.Serializable {
 			return anyObject
 		}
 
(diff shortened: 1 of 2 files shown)


Sources/Defaults/Utilities.swift
```diff
@@ -194,7 +194,7 @@ extension Defaults.Serializable {
 			return anyObject
 		}
 
-		guard let nextType = T.Serializable.self as? any Defaults.Serializable.Type else {
+        guard let nextType = T.Serializable.self as? any Defaults.Serializable.Type, nextType != T.self else {
```
hank121314, line 197: Please fix the indention, we are using tab indention.

Tests/DefaultsTests/DefaultsAnySeriliazableTests.swift
```diff
@@ -466,4 +466,10 @@ final class DefaultsAnySerializableTests: XCTestCase {
 
 		waitForExpectations(timeout: 10)
 	}
+
+    func testWrongCast() {
+        let value = Defaults.AnySerializable(false)
+        XCTAssertEqual(value.get(Bool.self), false)
+        XCTAssertNil(value.get(String.self))
+    }
```
hank121314, line 474: Same as the above comment, and please use `XCTAssertFalse` instead of `XCTAssertEqual` as linter suggests.

hank121314 (changes requested):
LGTM! Many thanks for the fix!
And please also fix the comments above. Thanks 🙇 !

Tests/DefaultsTests/DefaultsAnySeriliazableTests.swift
leoMehlig: I initially used that, but since `get` returns on optional, I'd have to provide a default value. That feels less clean.

Sources/Defaults/Utilities.swift
leoMehlig: Fixed that. Sorry about that.

PR #148, opened 2023-09-03
Adds `publisher` function, accepting array of keys

leoMehlig:
In my app I have a use case, where I have a dynamic array of keys, which I want to observe using a Publisher (a publisher works better here then AsyncSequence). By giving an option to use either variadic parameters or an array, that would be possible without breaking any code.

Additionally, I've discovered a crash when observing multiple keys and immediately cancelling the publisher. Here an observer, which was not yet added, was removed. Added check to make sure, this can't happen.

### DIFF
--- a/Sources/Defaults/Observation+Combine.swift
+++ b/Sources/Defaults/Observation+Combine.swift
@@ -107,7 +107,7 @@ extension Defaults {
 	) -> AnyPublisher<Void, Never> {
 		let initial = Empty<Void, Never>(completeImmediately: false).eraseToAnyPublisher()
 
-		let combinedPublisher =
+		return
 			keys
 				.map { key in
 					DefaultsPublisher(suite: key.suite, key: key.name, options: options)
@@ -117,7 +117,5 @@ extension Defaults {
 				.reduce(initial) { combined, keyPublisher in
 					combined.merge(with: keyPublisher).eraseToAnyPublisher()
 				}
-
-		return combinedPublisher
 	}
 }
--- a/Sources/Defaults/Observation+Combine.swift
+++ b/Sources/Defaults/Observation+Combine.swift
@@ -128,6 +128,6 @@ extension Defaults {
 		keys: _AnyKey...,
 		options: ObservationOptions = [.initial]
 	) -> AnyPublisher<Void, Never> {
-		self.publisher(keys: keys, options: options)
+		publisher(keys: keys, options: options)
 	}
 }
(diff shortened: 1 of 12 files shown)


Sources/Defaults/Observation.swift
```diff
@@ -138,10 +139,14 @@ extension Defaults {
 
 		func start(options: ObservationOptions) {
 			object?.addObserver(self, forKeyPath: key, options: options.toNSKeyValueObservingOptions, context: nil)
+			self.isObserving = true
```
sindresorhus, line 142:
```suggestion
			isObserving = true
```

Sources/Defaults/Observation.swift
```diff
@@ -138,10 +139,14 @@ extension Defaults {
...
 		}
 
 		func invalidate() {
-			object?.removeObserver(self, forKeyPath: key, context: nil)
+			if isObserving {
+				object?.removeObserver(self, forKeyPath: key, context: nil)
+				isObserving = false
+			}
```
sindresorhus, line 149:
```suggestion
			}

```

sindresorhus (changes requested):

Sources/Defaults/Observation+Combine.swift
```diff
@@ -118,4 +118,16 @@ extension Defaults {
...
+	/**
+	Publisher for multiple `Key<T>` observation, but without specific information about changes.
+
+	 - Warning: This method exists for backwards compatibility and will be deprecated sometime in the future. Use ``Defaults/updates(_:initial:)-9eh8`` instead.
+	*/
+	public static func publisher(
+		keys: _AnyKey...,
+		options: ObservationOptions = [.initial]
+	) -> AnyPublisher<Void, Never> {
+		self.publisher(keys: keys, options: options)
```
sindresorhus, line 131:
```suggestion
		publisher(keys: keys, options: options)
```

Sources/Defaults/Observation.swift
```diff
@@ -125,6 +125,7 @@ extension Defaults {
 		private weak var object: UserDefaults?
 		private let key: String
 		private let callback: Callback
+		private var isObserving = false
```
hank121314, line 128: How about using [observationInfo](https://developer.apple.com/documentation/objectivec/nsobject/1414009-observationinfo) to check whether the observer is registered?

leoMehlig: I'm uncertain if that works because the `observationInfo` is set on the object (`UserDefaults` in this case) and not on the observer. Therefore, even if the registration of this observer has not happened, another instance might be observing the `UserDefaults`, in which case `observationInfo` is non-nil.

hank121314: > another instance might be observing the UserDefaults, in which case observationInfo is non-nil.

True, thanks for the reminder!

hank121314 (approved):

PR #149, opened 2023-09-10
Fix support for non-string keys in dictionaries

sindresorhus:
`UserDefaults#set(dictionary, forKey: ...)` only supports a string-keyed dictionary. If you try to set `[1: true]`, it currently throws an Objective-C exception.


hank121314 (approved):
Ah! True! Many Thanks for the fix 🙌!

PR #151, opened 2023-10-04
Remove value wrapper from CodableBridge

### LINKED ISSUE #32: Remove value wrapper
We currently wrap all saved values in an array:

https://github.com/sindresorhus/Defaults/blob/55ffea9487fb9b559406d909ee31dcd955fe77aa/Sources/Defaults/Defaults.swift#L325-L327

This issue was fixed in macOS 10.15, iOS 13, etc, so it's not needed anymore on those platforms. However, we cannot remove it as it would require us to only support the latest OS versions. Instead, we can introduce a `Defaults.useLegacyWrapping` (please suggest better name) property that is `true` by default, but new projects that only target the latest OS versions could set this to true to get a slightly nicer output in UserDefaults.

sindresorhus:
Fixes #32

### DIFF
--- a/Sources/Defaults/Defaults+Bridge.swift
+++ b/Sources/Defaults/Defaults+Bridge.swift
@@ -12,13 +12,10 @@ extension Defaults.CodableBridge {
 		}
 
 		do {
-			// Some codable values like URL and enum are encoded as a top-level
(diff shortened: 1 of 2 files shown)


hank121314 (approved):
LGTM! Since `Defaults` is targeting macOS 10.15+.

PR #155, opened 2023-12-27
New init with default parameter that accepts another key

Maschina:
Proposal to create a key with another key as default value.
This can be useful if you want to change the name of the key, but still provide support to migrate from the previous key name.

### DIFF
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -153,6 +153,23 @@ extension Defaults {
 
 			super.init(name: name, suite: suite)
 		}
+		
+		/**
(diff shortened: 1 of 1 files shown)


sindresorhus:
It would be better if it accepted a string name as the `default` key (or `previousName` as a better name), since then you don't have to use two `Defaults.Key`. You just rename the existing `Defaults.Key` property name and then refer to the old UserDefaults name in the `previousName ` parameter. That being said, I'm not even sure this init carries its own weight as you can generally just keep the UserDefaults string name while renaming the `Defaults.Key` property. I do that all the time. Much easier.

Issue #144, opened 2023-07-02
MenuBarExtra freezes in a state loop when using Defaults


othyn:
It appears someone published a similar issue to this, but didn't respond when macOS 13 was retail released: https://github.com/sindresorhus/Defaults/issues/106

The example remains the same from that issue - in fact that issue is still directly happening exactly as described. With the issue Wouter01 was having being nearly identical in setup to mine.

Attempting to use `@Default` will break the app and lock it into a state loop:

```swift
@main
struct AutoClickerApp: App {
    @Default(.menuBarShowIcon) private var menuBarShowIcon

    var body: some Scene {
        MenuBarExtra(isInserted: $menuBarShowIcon,
                     content: { Label("Hi!") },
                     label: { Image(systemName: "cursorarrow.click.2") })
            .menuBarExtraStyle(.menu)
    }
}
```

However to provide more insights into the issue, here is the exact error message being given:

```
2023-07-03 00:27:25.238937+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.240590+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.244646+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.245899+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.249397+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.250887+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.255733+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
2023-07-03 00:27:25.257386+0100 AutoClicker[8113:47383804] [SwiftUI] Publishing changes from within view updates is not allowed, this will cause undefined behavior.
```

This will hang the app as soon as the main window loads and is pulled into focus, with the app entering a state loop, with the debug console being spammed with the above error message.

[...]

PR #158, opened 2024-01-18
Fix for freezes in MenuBarExtra and NavigationStack

Wouter01:
This PR changes the internal code of the `@Default` property wrapper, so it uses StateObject instead of ObservableObject. This solves issue #144 

As StateObject is not supported in SwiftUI 1.0, an `@available` has been added. This could be a breaking change for projects with the minimum target set to iOS 13.0 etc.

Instead of creating a new model when the key changes (which is what happened previously), the key will now be changed in the model itself. An extra check is done to make sure that the observers are not restarted too often.

### DIFF
--- a/Sources/Defaults/Defaults.swift
+++ b/Sources/Defaults/Defaults.swift
@@ -36,6 +36,8 @@ public enum Defaults {
 }
 
 public typealias _Defaults = Defaults
+
+@available(macOS 11, iOS 14, tvOS 14, watchOS 7, *)
 public typealias _Default = Default
 
 extension Defaults {
--- a/Sources/Defaults/SwiftUI.swift
+++ b/Sources/Defaults/SwiftUI.swift
@@ -6,14 +6,14 @@ extension Defaults {
 	final class Observable<Value: Serializable>: ObservableObject {
 		private var cancellable: AnyCancellable?
 		private var task: Task<Void, Never>?
-        
-        var key: Defaults.Key<Value> {
-            didSet {
-                if oldValue != key {
-                    observe()
-                }
-            }
-        }
+		
+		var key: Defaults.Key<Value> {
+			didSet {
+				if oldValue != key {
+					observe()
+				}
+			}
+		}
 
 		var value: Value {
 			get { Defaults[key] }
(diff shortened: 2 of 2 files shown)


Sources/Defaults/Defaults.swift
```diff
@@ -36,6 +36,8 @@ public enum Defaults {
 }
 
 public typealias _Defaults = Defaults
+
+@available(macOS 11, iOS 14, tvOS 14, watchOS 7, *)
```
sindresorhus, line 40: These are not needed.

https://github.com/sindresorhus/Defaults/commit/fcdf1967b967840862cd7a5bde5712b45bf79493

Sources/Defaults/SwiftUI.swift
```diff
@@ -6,25 +6,39 @@ extension Defaults {
 	final class Observable<Value: Serializable>: ObservableObject {
 		private var cancellable: AnyCancellable?
 		private var task: Task<Void, Never>?
-		private let key: Defaults.Key<Value>
-
-		let objectWillChange = ObservableObjectPublisher()
+		
+		var key: Defaults.Key<Value> {
+			didSet {
+				if oldValue != key {
```
sindresorhus, line 12:
```suggestion
				if key != oldValue {
```

Sources/Defaults/SwiftUI.swift
```diff
@@ -6,25 +6,39 @@ extension Defaults {
...
+					observe()
+				}
+			}
+		}
 
 		var value: Value {
 			get { Defaults[key] }
 			set {
-				objectWillChange.send()
```
sindresorhus, line 16: I don't see any harm in keeping this? Unless I'm missing something. See: https://github.com/sindresorhus/Defaults/issues/144#issuecomment-1873194410

sindresorhus (changes requested):

Sources/Defaults/SwiftUI.swift
Wouter01: If that is kept, the menubar will still freeze for some reason. Removing it fixes that, and doesn't seem to affect anything

sindresorhus (approved):

hank121314 (approved):
Using `@StateObject` looks good in this case, thanks for the fix!

sindresorhus:
https://github.com/sindresorhus/Defaults/releases/tag/v8.0.0

Sources/Defaults/SwiftUI.swift
sindresorhus: @Wouter01 Removing `objectWillChange.send()` broke animations when using `withAnimation`, so I will have to add it back unless you can think of a better solution.

Reproduction: [Fixture.zip](https://github.com/sindresorhus/Defaults/files/15469820/Fixture.zip)

PR #159, opened 2024-01-18
Remove v5 migration functionality

sindresorhus:
I plan a new major release, so I think it's time to remove it. I will make it clear in the release notes that users need to migrate to v5 first if they are coming from v4.


hank121314 (approved):
LGTM! Looking forward to the release!

Issue #170, opened 2024-04-20
Improve `UserDefaultsKeyObservation`


sindresorhus:
https://github.com/sindresorhus/Defaults/pull/136#discussion_r1543401152

PR #181, opened 2024-07-20
Summarize internal `UserDefaults` observation class

hank121314:
## Description

This PR fixes: #170.
Summarize all internal `UserDefaults` related observation classes into one.
Introduce a new class, `Defaults.UserDefaultsAnyKeysObservation`, which can observe multiple keys and can add or remove the keys being observed(useful with `Defaults.iCloud`).

### DIFF
--- a/Sources/Defaults/Observation.swift
+++ b/Sources/Defaults/Observation.swift
@@ -119,150 +119,11 @@ extension Defaults {
 		Thread.current.threadDictionary[key] = false
 	}
 
-	final class UserDefaultsKeyObservation: NSObject, Observation {
-		typealias Callback = (BaseChange) -> Void
-
-		private weak var object: UserDefaults?
-		private let key: String
-		private let callback: Callback
-		private var isObserving = false
-
-		init(object: UserDefaults, key: String, callback: @escaping Callback) {
-			self.object = object
-			self.key = key
-			self.callback = callback
-		}
-
-		deinit {
-			invalidate()
-		}
-
-		func start(options: ObservationOptions) {
-			object?.addObserver(self, forKeyPath: key, options: options.toNSKeyValueObservingOptions, context: nil)
-			isObserving = true
-		}
-
-		func invalidate() {
-			if isObserving {
-				object?.removeObserver(self, forKeyPath: key, context: nil)
-				isObserving = false
-			}
-
-			object = nil
-			lifetimeAssociation?.cancel()
-		}
-
-		private var lifetimeAssociation: LifetimeAssociation?
-
-		func tieToLifetime(of weaklyHeldObject: AnyObject) -> Self {
-			// swiftlint:disable:next trailing_closure
-			lifetimeAssociation = LifetimeAssociation(of: self, with: weaklyHeldObject, deinitHandler: { [weak self] in
-				self?.invalidate()
-			})
-
-			return self
-		}
-
-		func removeLifetimeTie() {
-			lifetimeAssociation?.cancel()
-		}
-
-		// swiftlint:disable:next block_based_kvo
-		override func observeValue(
-			forKeyPath keyPath: String?,
-			of object: Any?,
-			change: [NSKeyValueChangeKey: Any]?, // swiftlint:disable:this discouraged_optional_collection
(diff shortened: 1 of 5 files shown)


sindresorhus:
As outlined in https://github.com/sindresorhus/Defaults/pull/136#discussion_r1543408414, I was looking to have class that handles the minimum amount of logic for the "low-level" observation handling for a **single key**, and then rather have another class that handles the high-level logic with multiple keys. It's generally beneficial to wrap low-level/legacy code in a wrapper to make it easier to reason about.

hank121314:
> As outlined in [#136 (comment)](https://github.com/sindresorhus/Defaults/pull/136#discussion_r1543408414), I was looking to have class that handles the minimum amount of logic for the "low-level" observation handling for a **single key**, and then rather have another class that handles the high-level logic with multiple keys. It's generally beneficial to wrap low-level/legacy code in a wrapper to make it easier to reason about.

Got your point!
Yeah, it is really good to wrap low-level/legacy code in a wrapper. It also makes it possible for me to focus on the high-level logic 👍.
Many thanks for the suggestion!

Sources/Defaults/Observation.swift
```diff
@@ -119,95 +119,48 @@ extension Defaults {
...
-		private let key: String
-		private let callback: Callback
-		private var isObserving = false
+	final class SuiteKeyPair: Hashable {
+		weak var suite: NSObject?
+		let key: String
 
-		init(object: UserDefaults, key: String, callback: @escaping Callback) {
-			self.object = object
+		init(suite: NSObject, key: String) {
```
sindresorhus, line 126: Why not use the specific type?

```suggestion
		init(suite: UserDefaults, key: String) {
```

Sources/Defaults/Observation.swift
```diff
@@ -119,95 +119,48 @@ extension Defaults {
...
 
-	final class UserDefaultsKeyObservation: NSObject, Observation {
-		typealias Callback = (BaseChange) -> Void
-
-		private weak var object: UserDefaults?
-		private let key: String
-		private let callback: Callback
-		private var isObserving = false
+	final class SuiteKeyPair: Hashable {
+		weak var suite: NSObject?
```
sindresorhus, line 123:
```suggestion
		weak var suite: UserDefaults?
```

Sources/Defaults/Observation.swift
```diff
@@ -119,95 +119,48 @@ extension Defaults {
...
+	Standard observation for `Defaults`.
+	The only class which handle the low level observation.
+	*/
+	final class DefaultsObservation: NSObject {
+		typealias Callback = (SuiteKeyPair, BaseChange) -> Void
 
-		private weak var object: UserDefaults?
-		private let key: String
+		static var observationContext = 0
+		private weak var suite: NSObject?
```
sindresorhus, line 150:
```suggestion
		private weak var suite: UserDefaults?
```

Sources/Defaults/Observation.swift
```diff
@@ -216,17 +169,20 @@ extension Defaults {
...
-				isObserving = false
+			guard isObserving else {
+				return
 			}
-
-			object = nil
+			suite?.removeObserver(self, forKeyPath: name)
+			isObserving = false
+			suite = nil
 		}
```
sindresorhus, line 186: Unrelated to this specific pull request, but I wonder if this is thread-safe. It may be smart to put a lock on it?

sindresorhus (changes requested):

Sources/Defaults/Observation.swift
hank121314: Put a lock on it. Maybe we should consider to use AtomicBoolean after swift 6?

Sources/Defaults/Observation.swift
hank121314: Fixed! Thanks!

Sources/Defaults/Observation.swift
sindresorhus: > Maybe we should consider to use AtomicBoolean after swift 6?

👍

sindresorhus:
Looks great now 👍

PR #183, opened 2024-07-25
Fix build error for SE-0401

kitwtnb:
In `SE-0401` , which will take effect in Swift 6, types with `@StateObject` will no longer be MainActor.  
`@StateObject` must be referenced from MainActor or a build error will occur, so `@MainActor` is added.  
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0401-remove-property-wrapper-isolation.md


sindresorhus:
This will cause a warning on `update()` and I'm not yet sure how to fix that. I would have assumed `DynamicProperty` requirements would be marked `@MainActor`, but they are not.

PR #184, opened 2024-07-25
Fix build error for SE-0409

kitwtnb:
`SE-0409` , introduced in Swift 6, changes `import` from implicit `public import` to `internal import`.  
The `internal import` does not allow imports to be `public` and will result in build errors, so `public` has been added.  
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md

### DIFF
--- a/Sources/Defaults/Defaults+Bridge.swift
+++ b/Sources/Defaults/Defaults+Bridge.swift
@@ -1,4 +1,5 @@
-import SwiftUI
+public import Foundation
+public import SwiftUI
 #if os(macOS)
 import AppKit
 #else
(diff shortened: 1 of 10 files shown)


kitwtnb:
CI may need to build with Xcode 16 to resolve build errors. 🫤

sindresorhus:
Swift 6 does not come with this setting on by default, and also, the point of it is to prefer internal import as much as possible.

kitwtnb:
https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315#the-swift-6-migration-2  
> There are a number of previously-accepted upcoming features that will be enabled by default in Swift 6

The forum states that it is enabled by default, but where can I find that it is disabled by default?

sindresorhus:
I don't remember where I saw it, but you can see it in the Xcode beta:

Screenshot 2024-07-25 at 15 44 22

PR #185, opened 2024-08-05
Disable syncing from local if its default after `iCloud.add`

### LINKED ISSUE #172: Question on 9.0 beta iCloud Syncing
Currently I use 8.x with .updates to sync with iCloud manually. However, I find that when syncing with iCloud, it is better to use updates with initial: Bool to false. Or iCloud value will be erased easily by initial value that a new device's newly installed app.

So my question is, is the iCloud Syncing can ignore the initial value by default?

hank121314:
## Description

This PR fixes #172.

By default, `Defaults` will create a sync task that automatically detects the data source using timestamps. However, this might cause an issue if the device is newly installed, as it might accidentally push the default value to iCloud. It's also unnecessary to publish the default value to iCloud upon initialization, since all devices should already have that value.

## Implementation

Added a new function `isDefaultValue` for `Defaults.Key<Value>`. 
To detect if the locally stored value is a default value, I had to change `Defaults.iCloud.add` to use a generic packed parameter. 

## Question

This change introduces a new issue: we cannot support `Defaults.iCloud.add` with an array of `Defaults.Key<Value>` due to the lack of variadic parameter support. Is there any way to resolve this?

### DIFF
--- a/Sources/Defaults/Defaults+iCloud.swift
+++ b/Sources/Defaults/Defaults+iCloud.swift
@@ -82,7 +82,8 @@ extension Defaults {
 		/**
 		Add the keys to be automatically synced.
 		*/
-		// TODO: support array of Defaults.Key after swift 6 Pack iteration
+		// TODO: Support array of Defaults.Key after Swift 6 pack iteration is supported.
+		// https://github.com/sindresorhus/Defaults/pull/185#discussion_r1704464183
 		public static func add<each Value: Defaults.Serializable>(_ keys: repeat Defaults.Key<each Value>) {
 			repeat synchronizer.add(each keys)
 		}
--- a/Sources/Defaults/Defaults+iCloud.swift
+++ b/Sources/Defaults/Defaults+iCloud.swift
@@ -39,7 +39,7 @@ extension Defaults {
 
 	## Dynamically Toggle Syncing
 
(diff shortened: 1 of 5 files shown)


Sources/Defaults/Defaults+iCloud.swift
```diff
@@ -82,15 +82,8 @@ extension Defaults {
...
-		}
-
-		/**
-		Add the keys to be automatically synced.
-		*/
-		public static func add(_ keys: [Defaults.Keys]) {
-			synchronizer.add(keys)
+		public static func add<each Value: Defaults.Serializable>(_ keys: repeat Defaults.Key<each Value>) {
+			repeat synchronizer.add(each keys)
 		}
```
sindresorhus, line 87: Untested and would require Swift 6, but maybe:

```swift
public static func add<each Value: Defaults.Serializable>(_ keys: repeat [Defaults.Key<each Value>]) {
	for key in repeat each keys {
		add(key)
	}
}
```

sindresorhus (changes requested):

Sources/Defaults/Defaults+iCloud.swift
hank121314: Yeah, will add a `TODO` for it first!

sindresorhus:
Looks great. Thanks for fixing this 👍

PR #187, opened 2024-08-31
Add `constant` method to support SwiftUI previews and tests

MikhailGasanov:
### Description
This PR introduces a constant method to the Default extension, allowing developers to easily create default values that can be used in SwiftUI previews and unit tests. This addition simplifies the process of providing mock data and ensures that default values are consistent across different components of the app.

### Key Changes
- Added a constant(_:) method to the Default extension.
- The method creates a Default<Value> instance with a specified constant value, particularly useful for SwiftUI previews and testing scenarios.
- Updated documentation to include a usage example within a SwiftUI preview.

### Usage Example
Here’s how you can use the new constant method in a SwiftUI preview:
```swift
extension Defaults.Keys {
    static let hasUnicorn = Key<Bool>("hasUnicorn", default: false)
}

struct ContentView: View {
    @Default(.hasUnicorn) var hasUnicorn

    var body: some View {
        Toggle("Toggle", isOn: $hasUnicorn)
    }
}

#Preview {
	ContentView(hasUnicorn: .constant(true))
}
```
In this example, the constant method is used to set up a mock for `hasUnicorn` key, making it easy to visualize the view with a predefined value during development.

### Testing
- Unit tests have been added to verify that the constant(_:) method correctly initializes Default<Value> instances with the specified constant values.
- Manual testing has been performed to ensure that the method works seamlessly in SwiftUI previews.

### Additional Notes
- This new functionality is backward-compatible and doesn’t affect any existing code.
- It is particularly helpful for scenarios where you need to isolate and test UI components with predictable data.

### DIFF
--- a/Sources/Defaults/SwiftUI.swift
+++ b/Sources/Defaults/SwiftUI.swift
@@ -274,3 +274,35 @@ private struct ViewStorage<Value>: DynamicProperty {
 		self._valueBox = .init(wrappedValue: ValueBox(value()))
 	}
 }
+
+extension Default {
+	/**
+	A `Default` instance with a constant value.
+	
+	This is particularly useful for setting default values in SwiftUI previews.
+
+	- Parameter value: The constant value to be used as the default.
+	- Returns: A `Default<Value>` instance initialized with the specified constant value.
+
+	```swift
+	extension Defaults.Keys {
+		static let hasUnicorn = Key<Bool>("hasUnicorn", default: false)
+	}
+
+	struct ContentView: View {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for working on this, but I don't think this is the optimal way to solve this problem. `@State` and also `@Default` should generally be private, so you wouldn't be able to just pass in values, and changing views like that to accommodate preview is generally the wrong approach. In addition, if we were to add a `.constant()` thing, it should not actually write to `UserDefaults` at all.

For preview specifically, the right solution is to use https://developer.apple.com/documentation/SwiftUI/Previewable()

Issue #142, opened 2023-06-07
Support using `@Default` in `@Observable` classes


sindresorhus:
I think we can finally support `@Default` in observable classes.

`@Obserable` expands class properties to:

```swift
var order: Order? {
    get { 
        self.access(keyPath: \.order)
        return _order
    }
    set {
        self.withMutation(keyPath: \.order) {
            _order = newValue
        }
    }
}
```

So we could make a macro that could be attached a property and it would generate similar code, but also integrate with `Defaults`.

It would be nice if `@Default` could be reused, but I'm not sure whether it's possible.

I'm open to other ideas on how to make us able to integrate Defaults into `@Observable` classes.

There is no rush with this one. We will probably not do anything about it until after iOS 17 is out.

staticdreams:
Any updates on this? iOS17 is just around the corner and I'm in the process of migrating ``@ObservableObject`` to ``Observable``

sindresorhus:
This will unfortunately not happen soon unless someone contributes the functionality. Creating a macro is not easy, and the macro system is still in flux. Even the built-in `@AppStorage` does not yet work with `@Observable`.

For now, you could use either of these packages. They work with `Defaults` if you use the same key.

- https://github.com/zijievv/user-defaults-observation
- https://github.com/davidsteppenbeck/ObservableUserDefault

PR #189, opened 2024-09-15
Implement @ObservableDefaults macro

kevinrpb:
Hi there! Thought I'd pitch in and contribute to this great package. This addresses #142 by creating a new macro that can be used inside `@Observable` classes.

The macro is implemented in a new `DefaultsMacros` module. The decision to do so is based on the introduction of a new dependency on `swift-syntax`, and to keep the main module dependency-free (specially since the syntax package adds quite a bit to build times). Please, let me know if you'd rather move it to the main module.

Settings the PR as draft for early review, but there are at least a couple things I want to do to get it ready. Also, I haven't put much thought into the naming; advice on this is appreciated.

## TODO

- [ ] Figure out if `@ObservationIgnored` can be added automatically as part of the macro.
- [ ] Add docstring to the macro declaration.
- [ ] Add info on how to use the macro to the README.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,15 @@
+{
+  "originHash" : "1ba60ebd54db82e47e39bc8db179589187c069067eb0a8cd6ec19d2301c5abc4",
+  "pins" : [
+    {
+      "identity" : "swift-syntax",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/swiftlang/swift-syntax",
+      "state" : {
+        "revision" : "cb53fa1bd3219b0b23ded7dfdd3b2baff266fd25",
+        "version" : "600.0.0"
+      }
+    }
+  ],
+  "version" : 3
+}
--- a/Package.swift
+++ b/Package.swift
@@ -1,5 +1,6 @@
 // swift-tools-version:5.10
 import PackageDescription
+import CompilerPluginSupport
 
 let package = Package(
 	name: "Defaults",
@@ -16,17 +17,47 @@ let package = Package(
 			targets: [
 				"Defaults"
 			]
+		),
+		.library(
+			name: "DefaultsMacros",
+			targets: [
+				"DefaultsMacros"
+			]
 		)
+	],
+	dependencies: [
+		.package(url: "https://github.com/swiftlang/swift-syntax", from: "600.0.0")
 	],
 	targets: [
 		.target(
 			name: "Defaults",
 			resources: [.copy("PrivacyInfo.xcprivacy")]
 		),
+		.macro(
+			name: "DefaultsMacrosDeclarations",
+			dependencies: [
+				"Defaults",
+				.product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
+				.product(name: "SwiftCompilerPlugin", package: "swift-syntax")
+			]
+		),
+		.target(
+			name: "DefaultsMacros",
+			dependencies: ["Defaults", "DefaultsMacrosDeclarations"]
+		),
 		.testTarget(
 			name: "DefaultsTests",
 			dependencies: [
 				"Defaults"
+			]
+		),
+		.testTarget(
+			name: "DefaultsMacrosTests",
+			dependencies: [
+				"DefaultsMacros",
+				"DefaultsMacrosDeclarations",
+				.product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
+				.product(name: "SwiftSyntaxMacrosTestSupport", package: "swift-syntax"),
 			]
 		)
 	]
--- a/Sources/DefaultsMacros/ObservableDefaults.swift
+++ b/Sources/DefaultsMacros/ObservableDefaults.swift
@@ -1,0 +1,10 @@
+import Foundation
+
+import Defaults
+
+@attached(accessor, names: named(get), named(set))
+public macro ObservableDefaults<Value>(_ key: Defaults.Key<Value>) =
+	#externalMacro(
+		module: "DefaultsMacrosDeclarations",
+		type: "ObservableDefaultsMacro"
+	)
--- a/Sources/DefaultsMacrosDeclarations/ObservableDefaultsMacro.swift
+++ b/Sources/DefaultsMacrosDeclarations/ObservableDefaultsMacro.swift
@@ -1,0 +1,117 @@
+import SwiftCompilerPlugin
+import SwiftDiagnostics
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
+
+public struct ObservableDefaultsMacro: AccessorMacro {
+	public static func expansion(
+		of node: AttributeSyntax,
+		providingAccessorsOf declaration: some DeclSyntaxProtocol,
+		in context: some MacroExpansionContext
+	) throws -> [AccessorDeclSyntax] {
+		// Must be attached to a property declaration.
+		guard let variableDeclaration = declaration.as(VariableDeclSyntax.self) else {
+			throw ObservableDefaultsError.notAttachedToProperty
+		}
+
+		// Must be attached to a variable property (i.e. `var` and not `let`).
+		guard variableDeclaration.bindingSpecifier.tokenKind == .keyword(.var) else {
+			throw ObservableDefaultsError.notAttachedToVariable
+		}
+
+		// Must be attached to a single property.
+		guard variableDeclaration.bindings.count == 1, let binding = variableDeclaration.bindings.first else {
+			throw ObservableDefaultsError.notAttachedToSingleProperty
+		}
+
+		// Must not provide an initializer for the property (i.e. not assign a value).
+		guard binding.initializer == nil else {
+			throw ObservableDefaultsError.attachedToPropertyWithInitializer
+		}
+
+		// Must not be attached to property with existing accessor block.
+		guard binding.accessorBlock == nil else {
+			throw ObservableDefaultsError.attachedToPropertyWithAccessorBlock
+		}
+
+		// Must use Identifier Pattern.
+		// See https://swiftinit.org/docs/swift-syntax/swiftsyntax/identifierpatternsyntax
+		guard let pattern = binding.pattern.as(IdentifierPatternSyntax.self)?.identifier else {
+			throw ObservableDefaultsError.attachedToPropertyWithoutIdentifierProperty
+		}
+
+		// Must receive arguments
+		guard let arguments = node.arguments else {
+			throw ObservableDefaultsError.calledWithoutArguments
+		}
+
+		// Must be called with Labeled Expression.
+		// See https://swiftinit.org/docs/swift-syntax/swiftsyntax/labeledexprlistsyntax
+		guard let expressionList = arguments.as(LabeledExprListSyntax.self) else {
+			throw ObservableDefaultsError.calledWithoutLabeledExpression
+		}
+
+		// Must only receive one argument.
+		guard expressionList.count == 1, let expression = expressionList.first?.expression else {
+			throw ObservableDefaultsError.calledWithMultipleArguments
+		}
+
+		return [
+			#"""
+			get {
+				access(keyPath: \.\#(pattern))
+				return Defaults[\#(expression)]
+			}
+			"""#,
+			#"""
+			set {
+				withMutation(keyPath: \.\#(pattern)) {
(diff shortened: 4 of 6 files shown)


sindresorhus:
Thanks for working on this 🙏

sindresorhus:
> The decision to do so is based on the introduction of a new dependency on swift-syntax, and to keep the main module dependency-free (specially since the syntax package adds quite a bit to build times).

👍

kevinrpb:
I think this is now ready for review. Please, let me know about any issues/concerns.

BTW I renamed the macro to `@Default` so it is familiar. Let me know what you thing about this!

readme.md
```diff
@@ -207,6 +207,22 @@ Note that it's `@Default`, not `@Defaults`.
 
 You cannot use `@Default` in an `ObservableObject`. It's meant to be used in a `View`.
 
+#### `@Default` in `@Observable`
+
+With the `@Default` macro, you can use `Defaults` inside `@Observable` classes that use the [Observation](https://developer.apple.com/documentation/observation) framework. Doing so is as simple as importing `DefaultsMacros` as adding two lines to a property (note that adding `@ObservationIgnored` is needed to prevent clashes with `@Observable`):
+
+```swift
+import Defaults
+import DefailtsMacros
```
bgerm, line 216: Minor typo in the example, which should read `import DefaultsMacros`.

Thanks for the PR.  I'm new to SwiftUI and just tried this out.

sindresorhus:
I don't think the `@Default` naming is going to work. The moment you import `DefaultsMacros` in a place that uses the `@Default` property wrapper, those become macro usages too, which doesn't work in a `View`. I guess we can name it `ObservableDefault`.

sindresorhus:
The macro is also missing an important feature; it does not trigger updates to the model if a `Defaults` key is changed outside of the model. For example, if you set `Defaults[.foo] = true` somewhere in your code, the model will not be triggered. I think the user would expect the model to always be in sync with `Defaults`.

Sources/DefaultsMacros/Default.swift
```diff
@@ -0,0 +1,47 @@
+import Defaults
+import Foundation
+
+/**
+Attached macro that adds support for using ``Defaults`` in ``@Observable`` classes.
+
+- Important: To prevent issues with ``@Observable``, you need to also add ``@ObservationIgnored`` to the attached property.
+
+This macro adds accessor blocks to the attached property similar to those added by `@Observable`.
+
```
sindresorhus, line 10: Maybe add a warning that the use of macros will slow down build times.

kevinrpb:
Thank you for fixing the typos.

> I don't think the `@Default` naming is going to work. The moment you import `DefaultsMacros` in a place that uses the `@Default` property wrapper, those become macro usages too, which doesn't work in a `View`. I guess we can name it `ObservableDefault`.

Ah, this makes sense in hindsight. Will rename 👍🏽 

> The macro is also missing an important feature; it does not trigger updates to the model if a `Defaults` key is changed outside of the model. For example, if you set `Defaults[.foo] = true` somewhere in your code, the model will not be triggered. I think the user would expect the model to always be in sync with `Defaults`.

Yeah, this is indeed an important shortcoming. I've been fiddling to see if I could come up with something that is easy to use and works well. This is what I arrived at (expanded version), let me know what you think @sindresorhus:

```swift
@Observable
final class TestModelWithoutMacro {
	@ObservationIgnored
	var testValue: String {
		get {
			_ = _testValueObserver                // <--- Make sure the lazy var is initialized

			access(keyPath: \.testValue)
			return Defaults[keyProperty]
		}
		set {
			withMutation(keyPath: \.testValue) {
				Defaults[keyProperty] = newValue
			}
		}
	}

	// Use a lazy var so we can access self
	@ObservationIgnored
	private lazy var _testValueObserver: Defaults.Observation =
		Defaults.observe(.test) { keyChange in
			self.testValue = keyChange.newValue
		}
```

sindresorhus:
Yeah, that looks good.

Should be `Defaults.observe(Defaults.Keys.test, options: []) { keyChange in` so it doesn't trigger an initial update.

Sources/DefaultsMacrosDeclarations/ObservableDefaultMacro.swift
```diff
@@ -0,0 +1,117 @@
...
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
+
+public struct ObservableDefaultMacro: AccessorMacro {
+	public static func expansion(
+		of node: AttributeSyntax,
+		providingAccessorsOf declaration: some DeclSyntaxProtocol,
+		in context: some MacroExpansionContext
+	) throws -> [AccessorDeclSyntax] {
```
sindresorhus, line 12:
```suggestion
	) throws(ObservableDefaultMacroError) -> [AccessorDeclSyntax] {
```

And then you can remove `ObservableDefaultMacroError` from the throw statements.

kevinrpb: Ah! typed throws are here and I forgot about them. Neat.

kevinrpb:
OK... I'm a bit confused. When typing out the observation property, since it is lazy Swift doesn't complain about capturing `self`. However, when it's added via macro expansion it does?

I need to investigate why this happens... Any ideas?

image

sindresorhus:
Not sure.

kevinrpb:
Seems like the compiler is straight up crashing when generating a `lazy var` with a macro ([swiftlang/swift#74087](https://github.com/swiftlang/swift/issues/74087)). Xcode eats the error, but I can see it when using `swift test` in a terminal.

Will ask there and try to come up with other alternatives in the meantime.

kevinrpb:
While there's been no activity about the `lazy` properties in macros, I do have something else to comment on the PR.

I was fiddling around with other alternatives, and an issue came up that also applies when using the `lazy` property: if the `@Observable` class is marked as `Sendable`, `lazy` properties cause a warning in Swift 5 and an error in Swift 6 due to their inherent mutability. Basically, this means that any solution that involves a `var` will cause warnings and/or errors.

Will keep thinking about other ways, but I'm no longer sure it's possible to do with the current macro capabilities. I may be missing some clever trick to it.

sindresorhus:
> Basically, this means that any solution that involves a var will cause warnings and/or errors.

Use a [lock](https://github.com/sindresorhus/Defaults/blob/a89f799930c92a85aa04b72131849d46c0833ab3/Sources/Defaults/Utilities.swift#L252) and mark the variable with `@unchecked Sendable`.

kevinrpb:
> Use a [lock](https://github.com/sindresorhus/Defaults/blob/a89f799930c92a85aa04b72131849d46c0833ab3/Sources/Defaults/Utilities.swift#L252) and mark the variable with `@unchecked Sendable`.

Thank you for the suggestion, a lock may work. But did you mean to mark the `@Observable` class with `@unchecked Sendable`? Not sure there is a way to just mark one property.

We could generate the code below and warn users that if they need the model to be `Sendable` they need to either use `@unchecked` or mark it `@MainActor`, either would work. However, I'm worried that suggesting marking it as `@unchecked` may hide other possible errors and people complain about it. Do you think adding a warning in the README is a good enough mitigation?

```swift
/* This source generates the model below.

@Observable
final class TestModel: @unchecked Sendable {
	@ObservableDefault(Defaults.Keys.test)
	@ObservationIgnored
	var testValue: String
}
*/
@Observable
final class TestModel: @unchecked Sendable {
	@ObservationIgnored
	var testValue: String {
		get {
			// Ensure the observation is set up, using the lock to prevent data races
			_testValueLock.lock()
			_ = _testValueObservation
			_testValueLock.unlock()

			access(keyPath: \.testValue)
			return Defaults[Defaults.Keys.test]
		}
		set {
			withMutation(keyPath: \.testValue) {
				Defaults[Defaults.Keys.test] = newValue
			}
		}
	}

	@ObservationIgnored
	private let _testValueLock: Lock = .make()
	@ObservationIgnored
	private lazy var _testValueObservation: Defaults.Observation? = {
		Defaults.observe(Defaults.Keys.test, options: []) { [weak self] keyChange in
			self?.testValue = keyChange.newValue
		}
	}()
}
```

sindresorhus:
I was thinking there was a keyword for ignoring sendable, but what I meant was wrap it in an unchecked sendable box:

```swift
final class LockedSendableValue<Value>: @unchecked Sendable {
	private var _value: Value
	private let lock = Lock()

	init(_ value: @autoclosure @Sendable () throws -> Value) rethrows {
		self._value = try value()
	}

	func withValue<T: Sendable>(
		_ operation: @Sendable (inout Value) throws -> T
	) rethrows -> T {
		try self.lock.withLock {
			var value = self._value
			defer { self._value = value }
			return try operation(&value)
		}
	}
}
```

---

> But did you mean to mark the @Observable class with @unchecked Sendable? Not sure there is a way to just mark one property.

No

sindresorhus:
How about something like this? This also correctly destroys the event subscription when the class deinits:

```swift
extension Defaults.Keys {
	static let test = Key<String>("test", default: "")
}

@Observable
final class TestModelWithoutMacro {
	private enum AssociatedKeys {
		static var observer = 1
	}

	@ObservationIgnored
	var testValue: String {
		get {
			if objc_getAssociatedObject(self, &AssociatedKeys.observer) == nil {
				let cancellable = Defaults.publisher(.test)
					.sink { [weak self] in
						self?.testValue = $0.newValue
					}
				objc_setAssociatedObject(self, &AssociatedKeys.observer, cancellable, .OBJC_ASSOCIATION_RETAIN)
			}

			access(keyPath: \.testValue)
			return Defaults[.test]
		}
		set {
			withMutation(keyPath: \.testValue) {
				Defaults[.test] = newValue
			}
		}
	}
}
```

sindresorhus:
Friendly bump in case you forgot about this :)

kevinrpb:
@sindresorhus I spent some time (not a lot) and it seems like using the associated objects may work, so I'm trying to put that together.

I need to figure out a way to generate unique keys for each property so they don't collide, but after that it should be looking good.

kevinrpb:
Seems like you can just defined the keys as `Void?`, so I made the macro add a private property like such per needed key.

Thank you so much for the help with that! Let me know what you think of this!

sindresorhus:
This works great! Tested it in a real app. Nice work.

---

Only one problem left. When enabling Swift 6 mode, I get a warning:

> /var/folders/ng/1dc3svp96z36746xz4czb25r0000gn/T/swift-generated-sources/@__swiftmacro_5Plash8CatModelC7catName17ObservableDefaultfMp_.swift:1:20 Static property '_objcAssociatedKey_catName' is not concurrency-safe because it is nonisolated global shared mutable state

Sources/DefaultsMacrosDeclarations/ObservableDefaultMacro.swift
```diff
@@ -0,0 +1,191 @@
...
+	public static func expansion(
+		of node: SwiftSyntax.AttributeSyntax,
+		providingPeersOf declaration: some SwiftSyntax.DeclSyntaxProtocol,
+		in context: some SwiftSyntaxMacros.MacroExpansionContext
+	) throws -> [SwiftSyntax.DeclSyntax] {
+		let property = try propertyPattern(of: declaration)
+		let associatedKey = associatedKeyToken(for: property)
+
+		return [
+			"private static var \(associatedKey): Void?"
```
sindresorhus, line 69:
```suggestion
			"private nonisolated(unsafe) static var \(associatedKey): Void?"
```

This fixes the Swift 6 warning. It should be safe given that we only use it for its address.

kevinrpb: Oh my bad, for some reason I thought Swift 6 mode was enabled in the package by default.

I will update the macro, and thank you for pointing to the fix 😄

sindresorhus (approved):

sindresorhus:
@kevinrpb This looks good from my side. Maybe you want to take a last look over and approve this before I merge?

kevinrpb:
@sindresorhus thank you for pushing the change.

I tested it in my own app and it all seems to be working properly, so unless you have anything else you'd like me to change, I figure it's ready for the merge.

Thank you once again for taking the time to look through all of it!

sindresorhus:
Thanks for contributing 🙏

sindresorhus:
https://github.com/sindresorhus/Defaults/releases/tag/9.0.0

PR #191, opened 2024-09-30
Add variadic generics version of `Defaults.updates()`

sindresorhus:


Sources/Defaults/Defaults.swift
```diff
@@ -292,13 +292,64 @@ extension Defaults {
...
+			}
+
+			var observations = [DefaultsObservation]()
+
+			if initial {
+				continuation.yield(getCurrentValues())
+			}
+
+			for key in repeat (each keys) {
+				let observation = DefaultsObservation(object: key.suite, key: key.name) { _, _  in
```
sindresorhus, line 325: Note: I'm intentionally not using CompositeDefaultsObservation for simplicity, as it's not really needed here.

Sources/Defaults/Defaults.swift
```diff
@@ -292,13 +292,64 @@ extension Defaults {
...
+				}
+
+				observation.start(options: [])
+				observations.append(observation)
+			}
+
+			let immutableObservations = observations
+
+			continuation.onTermination = { _ in
+				// `invalidate()` should be thread-safe, but it is not in practice.
```
hank121314, line 336: Curious about this. I thought that `invalidate()` would be thread-safe with Lock, 
Should we use `DispatchQueue.sync` instead inside the `DefaultsObservation`?

sindresorhus: I'm not 100% sure. I have seen KVO crashes in some of my apps that may have been caused by some threading problems in macOS. I prefer to keep it limited to this code until I can 100% confirm it. This is just a precaution.

hank121314: Got it! Thanks for the explanation!

hank121314 (approved):

PR #209, opened 2025-07-29
Abort the synchronization task if no data source exists

### LINKED ISSUE #206: iCloud Settings Sync Issue: New App Installation Overwrites Existing Values with Empty Defaults
### Problem Description

When creating settings that sync through iCloud, I encounter the following issue:
1. Set some values in the app on Device A
2. Install the app on Device B (new installation)
3. The new app shows empty default values for the same settings
4. These empty values overwrite the existing iCloud data
5. All devices now show empty values instead of the original data

### Expected Behavior

- Wait for iCloud sync to complete before initializing settings
- Load existing values from iCloud instead of using local defaults
- Not overwrite existing iCloud data with empty defaults

hank121314:
## Description

This PR fixes #206.

Previously, when a user newly installed the app, we would create some synchronization tasks.  
Since the app was newly installed, both the remote and local data sources did not have a timestamp record, so we would sync the local data source to the remote one, which might override values in iCloud.

With this PR, we will abort the synchronization task if both the remote and local data sources do not have a timestamp record.  
This prevents unintended data overrides and allows developers to wait for iCloud synchronization instead (call `synchronize` on app launch and wait for `didChangeExternallyNotification`).

## Implementation

- Make `Defaults.iCloud.latestDataSource` return an optional. If it returns `nil`, we should abort the synchronization task.
- Add a new `SyncStatus.abort` in the logger to indicate that the key synchronization has been aborted.

## Discussion

Should we provide a function that ensures iCloud is synced?

ex. 

```swift
func waitForRemoteSynced() async -> Notification? {
	await NotificationCenter.default.notifications(named: NSUbiquitousKeyValueStore.didChangeExternallyNotification).first
}
```

### DIFF
--- a/Sources/Defaults/Defaults+iCloud.swift
+++ b/Sources/Defaults/Defaults+iCloud.swift
@@ -196,6 +196,7 @@ private enum SyncStatus {
 	case idle
 	case syncing
 	case completed
+	case aborted
 }
 
 /**
@@ -326,12 +327,18 @@ final class iCloudSynchronizer {
 
 	- Parameter keys: If the keys parameter is an empty array, the method will use the keys that were added to `Defaults.iCloud`.
 	- Parameter source: Sync keys from which data source (remote or local).
+
+	- Note: The synchronization task might be aborted if both the remote and local data sources do not exist.
 	*/
 	func syncWithoutWaiting(_ keys: [Defaults.Keys] = [], _ source: Defaults.iCloud.DataSource? = nil) {
 		let keys = keys.isEmpty ? Array(self.keys) : keys
 
 		for key in keys {
-			let latest = source ?? latestDataSource(forKey: key)
+			// If no data source is specified, we should abort this synchronization task.
+			guard let latest = source ?? latestDataSource(forKey: key) else {
+				Self.logKeySyncStatus(key, source: nil, syncStatus: .aborted, value: nil)
+				continue
+			}
 			enqueue {
 				await self.syncKey(key, source: latest)
 			}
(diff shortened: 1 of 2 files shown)


Sources/Defaults/Defaults+iCloud.swift
```diff
@@ -483,16 +490,23 @@ final class iCloudSynchronizer {
...
+		return switch (remoteTimestamp, localTimestamp) {
+		// If the local timestamp does not exist, use the remote timestamp as the latest data source.
+		case (.some(_), nil):
+			.remote
 		// If the remote timestamp does not exist, use the local timestamp as the latest data source.
-		guard let remoteTimestamp = timestamp(forKey: key, source: .remote) else {
-			return .local
+		case (nil, .some(_)):
+			.local
+		case let (.some(localTimestamp), .some(remoteTimestamp)):
```
sindresorhus, line 504: This is inverted.

Make sure to add a test for this case that would have failed.

Sources/Defaults/Defaults+iCloud.swift
```diff
@@ -326,12 +327,18 @@ final class iCloudSynchronizer {
...
+	- Note: The synchronization task might be aborted if both the remote and local data sources do not exist.
 	*/
 	func syncWithoutWaiting(_ keys: [Defaults.Keys] = [], _ source: Defaults.iCloud.DataSource? = nil) {
 		let keys = keys.isEmpty ? Array(self.keys) : keys
 
 		for key in keys {
-			let latest = source ?? latestDataSource(forKey: key)
+			// If no data source is specified, we should abort the synchronization task.
+			guard let latest = source ?? latestDataSource(forKey: key) else {
+				Self.logKeySyncStatus(key, source: .local, syncStatus: .abort, value: nil)
```
sindresorhus, line 339:
```suggestion
				Self.logKeySyncStatus(key, source: nil, syncStatus: .abort, value: nil)
```

?

Sources/Defaults/Defaults+iCloud.swift
```diff
@@ -193,6 +193,7 @@ extension Defaults.iCloud {
 }
 
 private enum SyncStatus {
+	case abort
```
sindresorhus, line 196:
```suggestion
	case aborted
```

?

And should be the last case.

sindresorhus (changes requested):

Sources/Defaults/Defaults+iCloud.swift
```diff
@@ -326,12 +327,18 @@ final class iCloudSynchronizer {
...
 	*/
 	func syncWithoutWaiting(_ keys: [Defaults.Keys] = [], _ source: Defaults.iCloud.DataSource? = nil) {
 		let keys = keys.isEmpty ? Array(self.keys) : keys
 
 		for key in keys {
-			let latest = source ?? latestDataSource(forKey: key)
+			// If no data source is specified, we should abort the synchronization task.
+			guard let latest = source ?? latestDataSource(forKey: key) else {
+				Self.logKeySyncStatus(key, source: .local, syncStatus: .abort, value: nil)
+				return
```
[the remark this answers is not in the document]
hank121314: Yeah, you're absolutely right, many thanks for catching this 🙏 .
Also add a test for it.

Sources/Defaults/Defaults+iCloud.swift
hank121314: My fault, thought it was already covered by the tests.
Added a test for `latestDataSource`.

